PR #36, opened 2026-05-12
feat(params): accept `--limitGenomeGenerateRAM` for STAR CLI compatibility

### LINKED ISSUE #25: `--limitGenomeGenerateRAM` rejected at CLI parser (STAR-compat regression)
## Summary

`rustar-aligner --runMode genomeGenerate` rejects the STAR-compatible `--limitGenomeGenerateRAM` flag at startup. STAR exposes this flag and any pipeline that wraps STAR derives a RAM cap from job resources and passes it through (e.g. the nf-core/rnaseq `STAR_GENOMEGENERATE` Nextflow module passes a value computed from `task.memory`).

## STAR reference behaviour

`--limitGenomeGenerateRAM <int>` is a documented STAR flag — see the [STAR manual](https://github.com/alexdobin/STAR/blob/master/doc/STARmanual.pdf) §3 — capping the resident-set used during genome generation. Default `31000000000` (~31 GB). It is independent from `--limitBAMsortRAM` (the BAM-sort memory cap).

## Reproducer

```bash
#!/usr/bin/env bash
set -euo pipefail
mkdir -p /tmp/rustar-mre-25 && cd /tmp/rustar-mre-25

echo "=== STAR with --limitGenomeGenerateRAM 31000000000 ==="
    --genomeDir /tmp/star-idx --genomeFastaFiles genome.fasta --sjdbGTFfile genes.gtf \
    --sjdbOverhang 100 --genomeSAindexNbases 7 \
    --limitGenomeGenerateRAM 31000000000 2>&1 | tail -3

echo
echo "=== rustar with the same flag ==="
    --genomeDir /tmp/rustar-idx --genomeFastaFiles genome.fasta --sjdbGTFfile genes.gtf \
    --sjdbOverhang 100 --genomeSAindexNbases 7 \
    --limitGenomeGenerateRAM 31000000000 2>&1 | tail -3
```

## Observed (verified on commit `5f8ad08` + STAR 2.7.11b)

STAR completes:

```
May 12 15:24:18 ... writing Suffix Array to disk ...
May 12 15:24:18 ... writing SAindex to disk
May 12 15:24:18 ..... finished successfully
```

rustar fails at the CLI parser, before any work:

```
error: unexpected argument '--limitGenomeGenerateRAM' found

  tip: a similar argument exists: '--limitBAMsortRAM'
```

## Suggested fix

In [`src/params.rs`](https://github.com/scverse/rustar-aligner/blob/main/src/params.rs) (near the existing `--limitBAMsortRAM` field around line 363), add:

```rust
#[arg(long = "limitGenomeGenerateRAM", default_value_t = 31_000_000_000_u64)]
pub limit_genome_generate_ram: u64,
```

Two acceptable behaviours, in order of preference:

1. **Accept and honour** — cap the resident-set during suffix-array / index construction (or at least the in-memory genome chunks).
2. **Accept and warn-ignore initially** (`log::warn!("--limitGenomeGenerateRAM accepted but not enforced yet; rustar uses its own memory management")`). This unblocks every STAR-compatible caller without committing to the cap implementation.

Either is dramatically better than failing at the CLI parser, because the flag isn't user-facing — it's emitted by every pipeline that wraps STAR.

## Why this matters

`rustar-aligner` positions itself as a STAR drop-in. Rejecting a flag STAR has accepted for years makes the drop-in claim conditional. nf-core/rnaseq's `STAR_GENOMEGENERATE` Nextflow module is one of many wrappers that will need a workaround otherwise.

---

pinin4fjords:
## Summary

Accepts the STAR-compatible `--limitGenomeGenerateRAM` flag at the CLI parser. The value is parsed and stored but not currently enforced — rustar's memory management is independent. This unblocks every wrapper that passes the flag derived from job resources (e.g. nf-core/rnaseq's `STAR_GENOMEGENERATE` Nextflow module).

## Before

```
$ rustar-aligner --runMode genomeGenerate --limitGenomeGenerateRAM 31000000000 ...
error: unexpected argument '--limitGenomeGenerateRAM' found
```

## After

```
$ rustar-aligner --runMode genomeGenerate --limitGenomeGenerateRAM 31000000000 ...
(proceeds; flag value accepted but not enforced)
```

## Test plan

- [x] `cargo build`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test --lib`
- [x] Smoke test confirms flag is no longer rejected at the CLI parser.

A future PR can wire the value into the suffix-array build path if real RAM capping is desired.

Fixes #25

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,6 +50,13 @@ fn genome_generate(params: &Parameters) -> anyhow::Result<()> {
             .collect::<Vec<_>>()
     );
 
+    if params.limit_genome_generate_ram != 31_000_000_000 {
+        log::warn!(
+            "--limitGenomeGenerateRAM {} accepted but not enforced; rustar manages genome-generation memory independently",
+            params.limit_genome_generate_ram
+        );
+    }
+
     info!("Building genome index...");
     let index = GenomeIndex::build(params)?;
 
--- a/src/params.rs
+++ b/src/params.rs
@@ -363,6 +363,12 @@ pub struct Parameters {
     #[arg(long = "limitBAMsortRAM", default_value_t = 0)]
     pub limit_bam_sort_ram: u64,
 
+    /// Maximum available RAM (bytes) for genome generation.
+    ///
+    /// Accepted for STAR CLI compatibility. Not currently enforced — rustar uses its own memory management.
+    #[arg(long = "limitGenomeGenerateRAM", default_value_t = 31_000_000_000_u64)]
+    pub limit_genome_generate_ram: u64,
+
     /// Route primary alignment output to stdout instead of a file.
     /// Values: None (default), SAM, BAM_Unsorted, BAM_SortedByCoordinate.
     #[arg(long = "outStd", default_value = "None")]
--- a/src/params.rs
+++ b/src/params.rs
@@ -363,9 +363,7 @@ pub struct Parameters {
     #[arg(long = "limitBAMsortRAM", default_value_t = 0)]
     pub limit_bam_sort_ram: u64,
 
-    /// Maximum available RAM (bytes) for genome generation.
-    ///
-    /// Accepted for STAR CLI compatibility. Not currently enforced — rustar uses its own memory management.
+    /// Maximum RAM (bytes) for genome generation.
     #[arg(long = "limitGenomeGenerateRAM", default_value_t = 31_000_000_000_u64)]
     pub limit_genome_generate_ram: u64,
 


pinin4fjords:
I'm unsure to what extent this makes sense really. I get Claude's rationale that it would be good not to break any CLI options though, even if they do nothing.

pinin4fjords:
Added a runtime warning when the flag is set to a non-default value. Verified on macOS/aarch64:

```
$ ./target/debug/rustar-aligner --runMode genomeGenerate --limitGenomeGenerateRAM 8000000000 ...
[WARN  rustar_aligner] --limitGenomeGenerateRAM 8000000000 accepted but not enforced; rustar manages genome-generation memory independently
Error: FASTA parsing error: no chromosomes found in FASTA files
```

Default and no-flag invocations stay silent (no spurious warning when the user didn't explicitly ask for a cap). A user passing a real RAM cap now sees that the cap isn't applied rather than having to read the PR body to find out.

Pushed as a follow-up commit (`cc3e48a`) on the same branch.

Psy-Fer:
I merged this, and extended it to take bytes and human readable values, like `64G`. So it's backward compatible, and forward "usable".

Psy-Fer:
merged in #77

Issue #26, opened 2026-05-12
`--outFileNamePrefix` ending in `.` is treated as a directory (STAR uses it as a literal string prefix)


pinin4fjords:
## Summary

When `--outFileNamePrefix` ends with `.` (or any path component without a separator), `rustar-aligner` treats it as a **directory name** and writes bare-named outputs inside it. STAR treats the same value as a **literal string prefix** and writes prefixed files at the top level.

This breaks every wrapper that emits files via STAR's prefix convention (`<sample>.Aligned.out.bam`, `<sample>.Log.final.out`, ...). We had to add a post-processing step in the nf-core/rnaseq integration to flatten the directory back to STAR-style filenames so the downstream collect-channels would work.

## STAR reference behaviour

STAR treats `--outFileNamePrefix` as a raw string prefix concatenated onto each output filename (no directory-component split). With `--outFileNamePrefix SAMPLE.` you get `SAMPLE.Aligned.out.bam`, `SAMPLE.Log.final.out`, `SAMPLE._STARpass1/` etc. at the top level — see [`source/Parameters.cpp`](https://github.com/alexdobin/STAR/blob/master/source/Parameters.cpp) where `outFileNamePrefix` is parsed and concatenated.

## Root cause in rustar

[`src/params.rs:346-347`](https://github.com/scverse/rustar-aligner/blob/main/src/params.rs#L346-L347):

```rust
#[arg(long = "outFileNamePrefix", default_value = "./")]
pub out_file_name_prefix: PathBuf,
```

Typing the prefix as `PathBuf` produces the directory-style behaviour: when the prefix ends in `.` (or `/`) it's joined with subsequent filenames as a path component rather than concatenated as a string.

## Reproducer

```bash
#!/usr/bin/env bash
set -euo pipefail
mkdir -p /tmp/rustar-mre-26 && cd /tmp/rustar-mre-26

BASE=https://raw.githubusercontent.com/nf-core/test-datasets/626c8fab639062eade4b10747e919341cbf9b41a
curl -fsLO $BASE/reference/genome.fasta
curl -fsL  $BASE/reference/genes_with_empty_tid.gtf.gz | gunzip -c > genes.gtf
curl -fsLO $BASE/testdata/GSE110004/SRR6357072_1.fastq.gz
curl -fsLO $BASE/testdata/GSE110004/SRR6357072_2.fastq.gz

RUSTAR=ghcr.io/scverse/rustar-aligner:dev
[...]

PR #37, opened 2026-05-12
fix(chimeric): create output parent directory before writing

### LINKED ISSUE #35: `--chimSegmentMin > 0` + `--twopassMode Basic` aborts the run when `--outFileNamePrefix` doesn't end in `/`
## Summary

`rustar-aligner --chimSegmentMin > 0` combined with `--twopassMode Basic` (and a `--outFileNamePrefix` value that doesn't end in `/`) aborts the run before writing any output:

```
[INFO  rustar_aligner] Chimeric detection enabled (chimSegmentMin=12)
Error: I/O error: No such file or directory (os error 2) (SAMPLE./Chimeric.out.junction)

Caused by:
    No such file or directory (os error 2)
```

No outputs are produced — `Aligned.out.bam`, `Log.final.out`, `SJ.out.tab`, and the chim file are all missing. The run is a hard failure.

This is a separable bug from [#26 (`--outFileNamePrefix` treated as a directory)](https://github.com/scverse/rustar-aligner/issues/26):

- **Without** `--twopassMode Basic`: rustar treats `<prefix>` as a directory and writes all outputs (including `Chimeric.out.junction`) inside it. The prefix-as-dir behaviour is the #26 bug, but it doesn't crash.
- **With** `--twopassMode Basic`: rustar still tries to write to `<prefix>/Chimeric.out.junction`, but in two-pass mode the chim writer fires in pass 1 *before* the dir-creating step that normal-output writing implicitly triggers, so the parent directory doesn't exist yet and the file open fails.

Pipelines that wrap STAR commonly add `--chimSegmentMin` via `extra_star_align_args` together with the rest of the standard STAR-fusion / arriba flag set, which includes `--twopassMode Basic`. Any nf-core/rnaseq run that enables both will crash silently (Nextflow surfaces only the non-zero exit; the error message points at a path issue, not a chim configuration issue).

## STAR reference behaviour

STAR concatenates `--outFileNamePrefix` as a literal string when emitting `<prefix>Chimeric.out.junction` — no directory split, no parent-dir requirement, works regardless of whether the prefix ends in `.`, `/`, or any other character. See [`source/Chimeric.cpp`](https://github.com/alexdobin/STAR/blob/master/source/Chimeric.cpp) and the file-opening glue in [`source/InOutStreams.cpp`](https://github.com/alexdobin/STAR/blob/master/source/InOutStreams.cpp). The output path is built once at startup from `P.outFileNamePrefix + "Chimeric.out.junction"` and lives next to the other STAR outputs.

## Reproducer

```bash
#!/usr/bin/env bash
set -euo pipefail
mkdir -p /tmp/rustar-mre-chim && cd /tmp/rustar-mre-chim

mkdir -p idx-rustar idx-star

# Identical flag set, two aligners. Both prefixes lack a trailing slash.
        --runThreadN 4 --sjdbGTFfile genes.gtf --twopassMode Basic --runRNGseed 0
        --outSAMtype BAM Unsorted --outSAMattributes NH HI AS NM MD
        --outSAMattrRGline ID:WT_REP2 SM:WT_REP2
        --quantMode TranscriptomeSAM --quantTranscriptomeSAMoutput BanSingleEnd
        --chimSegmentMin 12 --chimOutType Junctions)

echo "=== STAR (--outFileNamePrefix STAR.) ==="
    --genomeDir idx-star "${COMMON[@]}" --outFileNamePrefix STAR.
echo "--- STAR outputs ---"
ls STAR.*Chimeric.out.junction STAR.Aligned.out.bam STAR.Log.final.out 2>&1

echo
echo "=== rustar (--outFileNamePrefix RUS.) ==="
    --genomeDir idx-rustar "${COMMON[@]}" --outFileNamePrefix RUS. 2>&1 | tail -8 || true
echo "--- rustar outputs (none expected; run failed) ---"
ls RUS.* 2>&1 || true
ls RUS./ 2>&1 || true

echo
echo "=== rustar with trailing-slash prefix (workaround) ==="
mkdir -p RUSdir
    --genomeDir idx-rustar "${COMMON[@]}" --outFileNamePrefix RUSdir/ 2>&1 | tail -3
echo "--- rustar outputs ---"
ls RUSdir/
```

## Observed (verified, fresh paired run on commit `5f8ad08`)

STAR with `--outFileNamePrefix STAR.`:

```
STAR.Aligned.out.bam
STAR.Aligned.toTranscriptome.out.bam
STAR.Chimeric.out.junction              # <-- chim file at top level, prefixed
STAR.Log.final.out
STAR.Log.out
STAR.Log.progress.out
STAR.SJ.out.tab
STAR._STARgenome/
STAR._STARpass1/
```

rustar with `--outFileNamePrefix RUS.`:

```
[INFO  rustar_aligner] Reading paired-end from ...
[INFO  rustar_aligner] Chimeric detection enabled (chimSegmentMin=12)
Error: I/O error: No such file or directory (os error 2) (RUS./Chimeric.out.junction)

Caused by:
    No such file or directory (os error 2)

--- rustar outputs (none expected; run failed) ---
ls: cannot access 'RUS.*': No such file or directory
```

rustar with `--outFileNamePrefix RUSdir/` (parent pre-created):

```
Aligned.out.bam
Chimeric.out.junction       # <-- correct 14-column format
Log.final.out
SJ.out.tab
```

So the data layer is fine — `Chimeric.out.junction` content matches STAR's format when the path is reachable. Only the path-building / parent-dir-creation step is broken, and only when chim output combines with two-pass mode.

## Suggested fix

The chimeric-output writer appears to construct its path as `<outFileNamePrefix>/Chimeric.out.junction`, treating the prefix as a directory unconditionally — but in two-pass mode it fires before the directory has been created (in non-two-pass mode, an earlier output writer happens to create it first, masking the bug). Either:

1. Route chimeric output through the same path-builder used for the main outputs (`Aligned.out.bam`, `Log.final.out`, ...). Those handle the bare-prefix case today, since [#26 (`--outFileNamePrefix` treated as dir)](https://github.com/scverse/rustar-aligner/issues/26) causes them to write inside the dir but it does at least create the dir along the way.
2. Have the chim writer call `std::fs::create_dir_all(parent_of(target_path))` before opening the output file. One-line fix in the chim-emission path.

The cleanest fix is #1 plus the broader [#26](https://github.com/scverse/rustar-aligner/issues/26) fix (treat prefix as a string concatenation, matching STAR). Then `--outFileNamePrefix SAMPLE.` yields `SAMPLE.Chimeric.out.junction` next to `SAMPLE.Aligned.out.bam` etc., matching STAR byte-for-byte.

## Test plan

```bash
# After fix, both invocations should succeed and produce a chim file:
rustar-aligner --twopassMode Basic --chimSegmentMin 12 ... --outFileNamePrefix SAMPLE.
ls SAMPLE.Chimeric.out.junction   # currently does not exist; should after fix

rustar-aligner --twopassMode Basic --chimSegmentMin 12 ... --outFileNamePrefix dir/
ls dir/Chimeric.out.junction      # already works today; should still work
```

## Why this matters

`--chimSegmentMin` is the standard way users enable fusion / chimeric detection on top of STAR — every STAR-fusion, arriba, and STAR-derived chimeric protocol begins with that flag. Most STAR wrappers (nf-core/rnaseq's `extra_star_align_args`, the STAR-fusion Snakemake wrapper, raw fusion-detection scripts) combine it with `--twopassMode Basic` and a STAR-prefix convention `--outFileNamePrefix SAMPLE.`. Today, any such invocation crashes against rustar before any output is written, with an error that doesn't obviously point at chim configuration.

## Severity

Medium. nf-core/rnaseq doesn't currently pass `--chimSegmentMin` in its default test profile, so this isn't blocking the PR — but every user who adds it via `extra_star_align_args` will hit a silent failure mode that's hard to attribute without reading rustar source.

---

pinin4fjords:
## Summary

`rustar-aligner --chimSegmentMin > 0` combined with `--twopassMode Basic` aborts the run when `--outFileNamePrefix` doesn't end in `/`:

```
[INFO  rustar_aligner] Chimeric detection enabled (chimSegmentMin=12)
Error: I/O error: No such file or directory (os error 2) (SAMPLE./Chimeric.out.junction)
```

The chim writer constructs its path as `<prefix>/Chimeric.out.junction` (treating `--outFileNamePrefix` as a directory, per the existing behaviour tracked in #26). In two-pass mode the chim writer fires before any other output writer creates that directory, so the `File::create` call fails. Without two-pass mode the bug is masked because another output writer incidentally creates the dir first.

## Fix

Call `std::fs::create_dir_all` on the parent of the chim output path before opening the file. Minimal, scoped fix - does not change the existing prefix-as-dir behaviour (that's left to #26).

## Test plan

- [x] New unit test asserts the chim writer succeeds when the parent dir doesn't exist
- [x] Existing tests pass
- [x] `cargo build`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo fmt --check`

After the fix, `rustar-aligner --chimSegmentMin 12 --twopassMode Basic --outFileNamePrefix sample. ...` completes and writes `sample./Chimeric.out.junction` (with 14-column content matching STAR's format).

Fixes #35

### DIFF
--- a/src/chimeric/output.rs
+++ b/src/chimeric/output.rs
@@ -295,10 +295,6 @@ mod tests {
 
     #[test]
     fn test_chimeric_junction_writer_creates_missing_parent() {
-        // Regression test for #35: when --twopassMode Basic is combined with
-        // --chimSegmentMin > 0 and --outFileNamePrefix doesn't end in `/`, the
-        // chim writer fires before any other output writer creates the parent
-        // directory. The writer must create the parent itself.
         let dir = tempdir().unwrap();
         let prefix_path = dir.path().join("sample.");
         let prefix = prefix_path.to_str().unwrap();
--- a/src/chimeric/output.rs
+++ b/src/chimeric/output.rs
@@ -25,6 +25,12 @@ impl ChimericJunctionWriter {
         let mut path = PathBuf::from(prefix);
         path.push("Chimeric.out.junction");
 
+        if let Some(parent) = path.parent()
+            && !parent.as_os_str().is_empty()
+        {
+            std::fs::create_dir_all(parent).map_err(|e| Error::io(e, parent))?;
+        }
+
         let file = File::create(&path).map_err(|e| Error::io(e, &path))?;
 
         let writer = BufWriter::new(file);
@@ -285,6 +291,30 @@ mod tests {
         let mut path = PathBuf::from(prefix);
         path.push("Chimeric.out.junction");
         assert!(path.exists());
+    }
+
+    #[test]
+    fn test_chimeric_junction_writer_creates_missing_parent() {
+        // Regression test for #35: when --twopassMode Basic is combined with
+        // --chimSegmentMin > 0 and --outFileNamePrefix doesn't end in `/`, the
+        // chim writer fires before any other output writer creates the parent
+        // directory. The writer must create the parent itself.
+        let dir = tempdir().unwrap();
+        let prefix_path = dir.path().join("sample.");
+        let prefix = prefix_path.to_str().unwrap();
+
+        assert!(!prefix_path.exists(), "parent dir should not exist yet");
+
+        let writer = ChimericJunctionWriter::new(prefix);
+        assert!(
+            writer.is_ok(),
+            "writer should create missing parent dir, got: {:?}",
+            writer.err()
+        );
+
+        let mut path = PathBuf::from(prefix);
+        path.push("Chimeric.out.junction");
+        assert!(path.exists(), "chim output file should exist at {:?}", path);
     }
 
     #[test]


pinin4fjords:
Verified the fix end-to-end on macOS/aarch64 with the binary built from `8589859`.

Ran the exact failure mode from the issue (PE + `--twopassMode Basic` + `--chimSegmentMin 12` + bare-dot `--outFileNamePrefix RUS.`) against the yeast test inputs:

```
[INFO  rustar_aligner] Writing pass 1 junctions to RUS./SJ.pass1.out.tab
[INFO  rustar_aligner] Pass 1 discovered 16 novel junctions
...
[INFO  rustar_aligner::stats] Total mapped: 42631 (85.26%)
[INFO  rustar_aligner] Wrote RUS./Log.final.out
[INFO  rustar_aligner] Alignment complete!

$ ls RUS./
Aligned.out.bam
Aligned.toTranscriptome.out.bam
Chimeric.out.junction
Log.final.out
SJ.out.tab
SJ.pass1.out.tab
```

Pre-fix the same invocation crashed before any output was written:

```
[INFO  rustar_aligner] Chimeric detection enabled (chimSegmentMin=12)
Error: I/O error: No such file or directory (os error 2) (RUS./Chimeric.out.junction)
```

After the fix the run completes cleanly, the parent dir is created by `create_dir_all`, and `Chimeric.out.junction` lands alongside the other outputs. LGTM from the consumer side.

src/chimeric/output.rs
```diff
@@ -24,6 +24,12 @@ impl ChimericJunctionWriter {
     pub fn new(prefix: &str) -> Result<Self, Error> {
         let path = PathBuf::from(format!("{prefix}Chimeric.out.junction"));
 
+        if let Some(parent) = path.parent()
+            && !parent.as_os_str().is_empty()
```
flying-sheep, line 28: What’s this for? How can a path have `""` as parent?

pinin4fjords: `Path::parent()` returns `Some("")` (not `None`) for single-component relative paths. The chim path is built as `format!("{prefix}Chimeric.out.junction")` with no `/` inserted, so a prefix like `sample.` gives `sample.Chimeric.out.junction`, whose parent is `""`:

```rust
PathBuf::from("sample.Chimeric.out.junction").parent()
// Some("")

PathBuf::from("out/sample.Chimeric.out.junction").parent()
// Some("out")
```

Without the guard we'd call `create_dir_all("")` for bare-relative prefixes. On current stable that's a no-op `Ok(())`, so the check is defensive rather than load-bearing. Happy to drop it if you'd rather lean on that.

flying-sheep: right, says so right in the `parent()` docs, I guess I’m bad at reading.

flying-sheep (approved):

Psy-Fer:
merged in #77

PR #46, opened 2026-05-12
fix(params): treat `--outFileNamePrefix` as a literal string prefix

### LINKED ISSUE #26: `--outFileNamePrefix` ending in `.` is treated as a directory (STAR uses it as a literal string prefix)

pinin4fjords:
## Summary

`--outFileNamePrefix SAMPLE.` (or any value not ending in `/`) was treated as a directory by rustar - outputs landed in `SAMPLE./Aligned.out.bam` etc. STAR concatenates the prefix straight onto each output filename: `SAMPLE.Aligned.out.bam` at the top level. Every STAR wrapper assumes the latter convention; the directory behaviour breaks the output-glob convention nf-core/rnaseq and most STAR Snakemake wrappers rely on.

Root cause: `out_file_name_prefix` was typed as `PathBuf` and joined via `.push("Aligned.out.bam")`. That always treats the existing value as a directory component.

## Fix

Change the type to `String` and concatenate at every output-write site:

```rust
// before:
let mut path = PathBuf::from(&params.out_file_name_prefix);
path.push("Aligned.out.bam");

// after:
let path = PathBuf::from(format!("{}Aligned.out.bam", params.out_file_name_prefix));
```

A small `Parameters::output_path("Aligned.out.bam")` helper does this concatenation once so the call sites stay readable. The chimeric junction writer (which already took the prefix as `&str`) was using `PathBuf::push` internally and is fixed the same way.

Trailing-slash prefixes (`out/`) still work - `format!` produces `out/Aligned.out.bam`, a valid path. The bare-dot case now produces `SAMPLE.Aligned.out.bam` at the top level, matching STAR.

## Test plan

- [x] New unit tests on `Parameters::output_path` cover bare-dot, trailing-slash, default, and path-with-underscore prefixes
- [x] New integration test `test_bare_dot_prefix_is_literal_string` runs the binary with `<run_dir>/SAMPLE.` and asserts `SAMPLE.Aligned.out.bam` + `SAMPLE.Log.final.out` land at the top level and that no `SAMPLE./` directory is created
- [x] New chimeric writer unit test `test_chimeric_junction_writer_bare_dot_prefix` covers the same case for `Chimeric.out.junction`
- [x] Existing integration tests (`tests/alignment_features.rs`, `tests/phase9_threading.rs`, `tests/transcriptome_sam.rs`) all already use trailing-slash prefixes and continue to pass unchanged
- [x] `cargo build`
- [x] `cargo clippy --lib -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test` - all 402 tests pass (388 lib + 9 alignment_features + 4 phase9_threading + 1 transcriptome_sam)

After this fix, no downstream wrapper needs the workaround that flattens `SAMPLE./` back to STAR-style filenames after the run.

Fixes #26

### DIFF
--- a/src/chimeric/output.rs
+++ b/src/chimeric/output.rs
@@ -22,8 +22,7 @@ impl ChimericJunctionWriter {
     ///
     /// Creates file: {prefix}Chimeric.out.junction
     pub fn new(prefix: &str) -> Result<Self, Error> {
-        let mut path = PathBuf::from(prefix);
-        path.push("Chimeric.out.junction");
+        let path = PathBuf::from(format!("{prefix}Chimeric.out.junction"));
 
         let file = File::create(&path).map_err(|e| Error::io(e, &path))?;
 
@@ -277,22 +276,38 @@ mod tests {
     #[test]
     fn test_chimeric_junction_writer_creation() {
         let dir = tempdir().unwrap();
-        let prefix = dir.path().to_str().unwrap();
+        let prefix = format!("{}/", dir.path().display());
 
-        let writer = ChimericJunctionWriter::new(prefix);
+        let writer = ChimericJunctionWriter::new(&prefix);
         assert!(writer.is_ok());
 
-        let mut path = PathBuf::from(prefix);
-        path.push("Chimeric.out.junction");
+        let path = PathBuf::from(format!("{prefix}Chimeric.out.junction"));
         assert!(path.exists());
+    }
+
+    #[test]
+    fn test_chimeric_junction_writer_bare_dot_prefix() {
+        let dir = tempdir().unwrap();
+        let prefix = format!("{}/SAMPLE.", dir.path().display());
+
+        let writer = ChimericJunctionWriter::new(&prefix);
+        assert!(writer.is_ok());
+
+        let path = PathBuf::from(format!("{prefix}Chimeric.out.junction"));
+        assert!(path.exists(), "expected {} to exist", path.display());
+        assert!(
+            path.file_name().unwrap().to_str().unwrap() == "SAMPLE.Chimeric.out.junction",
+            "expected literal concatenation, got {}",
+            path.display()
+        );
     }
 
     #[test]
     fn test_write_inter_chromosomal() {
         let dir = tempdir().unwrap();
-        let prefix = dir.path().to_str().unwrap();
+        let prefix = format!("{}/", dir.path().display());
 
-        let mut writer = ChimericJunctionWriter::new(prefix).unwrap();
+        let mut writer = ChimericJunctionWriter::new(&prefix).unwrap();
 
         // Create mock chimeric alignment (chr9 -> chr22, BCR-ABL fusion)
         let donor = ChimericSegment {
@@ -337,8 +352,7 @@ mod tests {
         writer.flush().unwrap();
 
         // Read file and verify
-        let mut path = PathBuf::from(prefix);
-        path.push("Chimeric.out.junction");
+        let path = PathBuf::from(format!("{prefix}Chimeric.out.junction"));
 
         let mut content = String::new();
         File::open(&path)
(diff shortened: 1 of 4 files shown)


pinin4fjords:
Verified end-to-end on macOS/aarch64 against the rebuilt fix branch.

**Bare-dot prefix** (`--outFileNamePrefix SAMPLE.`) - the issue's failure case:

```
$ ls SAMPLE.*
SAMPLE.Aligned.out.bam
SAMPLE.Log.final.out
SAMPLE.SJ.out.tab
SAMPLE.SJ.pass1.out.tab

$ [ -d "SAMPLE." ] && echo dir-exists || echo no-dir
no-dir
```

Top-level `SAMPLE.Aligned.out.bam`, `SAMPLE.Log.final.out`, `SAMPLE.SJ.out.tab` all land, matching STAR's literal-string-prefix convention. No `SAMPLE./` directory created.

**Trailing-slash prefix** (`--outFileNamePrefix outdir/`) - regression check:

```
$ ls outdir/
Aligned.out.bam
Log.final.out
SJ.out.tab
SJ.pass1.out.tab
```

Still produces files inside the directory as expected; the fix didn't regress the directory-style usage.

LGTM.

PR #45, opened 2026-05-12
fix(sjdb): key `SpliceJunctionDb` in genome-absolute 0-based coordinates

pinin4fjords:
## Summary

`SpliceJunctionDb` was keyed in chromosome-local 1-based coordinates (extracted straight from GTF column 5) but consulted in two **different** coordinate spaces at runtime:

| Call site | Convention | On chr 0 (chr_start=0) | On chr N+ |
|---|---|---|---|
| stitch-time (`src/align/stitch.rs:1305-1314`) | genome-absolute 0-based | off by 1 -> miss | off by `chr_start[N]` -> miss |
| stats-time (`src/lib.rs:1860-1894`)           | genome-absolute 1-based  | matches -> hit  | off by `chr_start[N]` -> miss |

Neither matched the DB. Smoking gun: on the same WT_REP2 BAM, `SJ.out.tab` had 2 rows `annotated=1` (stats-time accidentally hits chr-0) while `Log.final.out` reported `Number of splices: Annotated (sjdb) | 0` (stitch-time always misses). The stitch-time miss is the load-bearing one - it drops `sjdb_score` and pulls in the stricter `align_sj_overhang_min` gate, costing ~50 % of GT/AG splices on the nf-core/rnaseq test profile.

## Fix

Normalise the DB to **genome-absolute 0-based** at construction (matching the existing convention used by `prepared_junctions` in `src/index/mod.rs` and by `SpliceJunctionStats`). Single source of truth contained in the GTF extraction step.

- `extract_junctions_configured` in `src/junction/gtf.rs` now adds `chr_start[chr_idx]` and subtracts 1 before returning each `(intron_start, intron_end)`. `SpliceJunctionDb::from_raw_junctions` consumes those tuples verbatim.
- Stats-time call sites at `src/lib.rs:1860` and `src/lib.rs:796` now record `(genome_pos, genome_pos + intron_len - 1)` instead of `(genome_pos + 1, genome_pos + intron_len)`. `genome_pos` is the 0-based first intronic base, which is what `detect_splice_motif` already expects.
- Stitch-time call site at `src/align/stitch.rs` is unchanged - it already passes genome-absolute 0-based.
- `SJ.out.tab` writer in `src/junction/sj_output.rs` now adds `+ 1` after subtracting `chr_start[chr_idx]` so the chr-local 1-based output is unchanged.
[...]

PR #51, opened 2026-05-12
fix: decode Gsj-region SA hits into split splice seeds

### LINKED ISSUE #47: pass-1 alignment doesn't seed candidates from `--sjdbGTFfile`; ~50% of GT/AG splices missed even after #27
## Summary

Pass-1 alignment doesn't use GTF-derived junctions as alignment candidates. Even after [#27](https://github.com/scverse/rustar-aligner/issues/27) (annotation lookup at scoring time) lands, total splice count stays around half STAR's on the same data, with the non-canonical rate elevated — rustar is only discovering junctions from the reads themselves, where STAR also seeds pass-1 from the GTF-derived `Gsj` buffer baked into the SA-extended genome at index time.

| `Log.final.out` field | rustar | STAR 2.7.11b |
|---|---:|---:|
| `Number of splices: Total`              | 366 | 801 |
| `Number of splices: Annotated (sjdb)`   | 0 (95 after #27) | 620 |
| `Number of splices: GT/AG`              | 266 | 686 |
| `Number of splices: Non-canonical`      | 93  | 108 |

Read alignments that would extend across a GTF junction fall through to the stricter unannotated gate and either map as unspliced (CIGAR collapse to `101M`) or fall to `Reads unmapped: too short`. Per-read example, fragment `SRR6357072.32572100`:

- STAR: `96M14674N5M` crossing one of the GTF-annotated yeast introns. NH=2.
- rustar: `101M` at the same exonic start. NH=1.

## STAR reference behaviour

STAR's pass-1 uses the sjdb-extended genome — the `Gsj` buffer appended to the genome and indexed in the SA during `genomeGenerate` (see [`source/sjdbPrepare.cpp`](https://github.com/alexdobin/STAR/blob/master/source/sjdbPrepare.cpp) and [`source/sjdbBuildIndex.cpp`](https://github.com/alexdobin/STAR/blob/master/source/sjdbBuildIndex.cpp)). Seeds that land in the `Gsj` region get translated back to the corresponding genome donor/acceptor pair and the splice becomes a candidate junction without the read having to bridge the gap itself.

rustar's `src/junction/sjdb_insert.rs` ports the sjdb-into-`Gsj` logic at `genomeGenerate` time, and the startup log confirms the junctions are loaded:

```
[INFO  rustar_aligner::junction] Extracted 5 annotated junctions from GTF
[INFO  rustar_aligner::index::io] Junction database loaded: 5 annotated junctions
```

So the index machinery is there. The gap is at align time: SA hits in the `Gsj` region are dropped at `Genome::position_to_chr` (returns `None` for positions in the Gsj buffer) instead of being decoded into splice candidates that propagate into the seed → cluster → stitch pipeline.

## Reproducer

```bash
#!/usr/bin/env bash
set -euo pipefail
mkdir -p /tmp/rustar-mre-47 && cd /tmp/rustar-mre-47

mkdir -p idx-rustar idx-star

        --runThreadN 4 --sjdbGTFfile genes.gtf --twopassMode Basic --runRNGseed 0
        --outSAMtype BAM Unsorted --outSAMattributes NH HI AS NM MD
        --outSAMattrRGline ID:WT_REP2 SM:WT_REP2)

    --genomeDir idx-rustar "${COMMON[@]}" --outFileNamePrefix RUS.
    --genomeDir idx-star "${COMMON[@]}" --outFileNamePrefix STAR.

echo "=== rustar splices ==="; grep -E 'splices' RUS./Log.final.out
echo "=== STAR splices ==="   ; grep -E 'splices' STAR.Log.final.out
```

## Suggested fix

At alignment time, SA hits in the `[n_genome_real, n_genome)` region (i.e. inside `Gsj`) need to be decoded back into `(donor_fwd, acceptor_fwd)` genome positions and added as splice candidates feeding the seed → cluster → stitch DP. Pseudocode mirroring STAR's `source/seedSearchStart.cpp` / `source/Aligner.cpp`:

```
for each SA hit:
    if hit_pos >= n_genome_real:
        (donor, acceptor) = decode_gsj_hit(hit_pos)
        emit splice_candidate(donor, acceptor)
```

Concretely: add a `decode_gsj_hit` helper in `src/junction/sjdb_insert.rs` (or wherever the Gsj layout is owned), expose `n_genome_real` on `Genome`, and route hits through the helper in `cluster_seeds` / equivalent.

## Severity

Medium. Mapping-rate impact on the yeast test profile is ~0.2 pp (~50 of 49 551 reads per sample shift from unique → unmapped). Per-read CIGAR impact is bigger: 234 of 340 same-position-different-CIGAR reads on `WT_REP2` are this pattern. Scales with how many reads in the input cross GTF junctions — yeast has 1.5 %, mammalian samples much more.

---

<sub>Filed during nf-core/rnaseq integration testing ([nf-core/rnaseq#1855](https://github.com/nf-core/rnaseq/pull/1855)).</sub>

pinin4fjords:
## Summary

Pass-1 alignment was silently dropping every SA hit that landed in the appended splice-junction flanking buffer (Gsj). Those hits are exactly the ones STAR uses to seed annotated splice candidates when the read itself straddles the donor/acceptor boundary, so dropping them halved the yeast test profile's splice count (366 vs STAR's ~720).

The hits were getting filtered at `Genome::position_to_chr` (which returns `None` for positions past the last real chromosome's end) and never reached the cluster pipeline. STAR handles them via the `g1 >= P.nGenomeReal` branch in `seedSearchStart.cpp` / `Aligner.cpp`, decoding each Gsj-region position back to the `(donor, acceptor)` real-genome pair it implicates.

## What changed

- `Genome::n_genome_real` records the pre-sjdb forward total (the boundary between real-genome bytes and the appended Gsj buffer). Set during `from_fasta` and on the load path; preserved across `append_sjdb`.
- `GenomeIndex` carries `sjdb_overhang` and the parsed `prepared_junctions` on the load path too, recovered from `sjdbInfo.txt`. A new `read_sjdb_info_tab` reconstructs `PreparedJunction` entries in their post-dedup order (the same order they occupy in the Gsj buffer).
- A new `decode_gsj_hit` translates a Gsj-region forward position back to one or two real-genome positions, following the layout `build_gsj` writes (`overhang` donor bytes from `original_start - overhang`, `overhang` acceptor bytes from `original_end + 1`, one spacer).
- `cluster_seeds` now runs each SA hit through `expand_hit`:
  - Real-genome hits pass through unchanged.
  - Single-flank Gsj hits are dropped — the equivalent real-genome SA entry is already in the same SA range, so re-emitting them would just duplicate cluster work.
  - Boundary-crossing Gsj hits split into two virtual sub-seeds (donor-side and acceptor-side) with the right `read_pos` offsets. The existing stitch DP chains them via its splice branch.
- Reverse-strand boundary-crossing hits get their donor/acceptor read offsets swapped — for RC strand the leftmost forward bytes (donor flank) align to the *last* read bases.

## Verification

Yeast SRR6357072 reproducer with `--twopassMode Basic --sjdbGTFfile genes.gtf --sjdbOverhang 100`:

| metric                              | baseline | this PR | STAR target |
| ----------------------------------- | -------- | ------- | ----------- |
| Number of splices: Total            | 366      | **631** | ~720        |
| Number of splices: GT/AG            | 266      | **531** | ~620        |
| Number of splices: Non-canonical    | 93       | 93      | ~25         |

`Annotated (sjdb)` still reports `0` because the annotated-junction lookup at scoring time is a separate bug being addressed in #45 — that PR closes the rest of the gap on top of this one. Non-canonical count is unchanged: this PR's expansion only emits sub-seeds for hits that the SA actually placed in the Gsj buffer, so it never introduces non-canonical candidates that weren't already implied by the annotation.

## Test plan

- [x] `cargo build` clean
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --lib -- -D warnings` clean
- [x] `cargo test --lib` (390 passed, 0 failed; 7 new tests cover `decode_gsj_hit` real-genome / single-flank / boundary-crossing / multi-junction / non-canonical-shift / out-of-bounds cases and `read_sjdb_info_tab` round-trip).
- [x] End-to-end yeast reproducer numbers above.

Fixes #47

### DIFF
--- a/src/align/read_align.rs
+++ b/src/align/read_align.rs
@@ -1505,6 +1505,7 @@ mod tests {
         let genome = Genome {
             sequence,
             n_genome,
+            n_genome_real: n_genome,
             n_chr_real: 1,
             chr_name: vec!["chr1".to_string()],
             chr_length: vec![10],
@@ -1535,6 +1536,7 @@ mod tests {
             junction_db: crate::junction::SpliceJunctionDb::empty(),
             transcriptome: None,
             prepared_junctions: Vec::new(),
+            sjdb_overhang: 0,
         }
     }
 
--- a/src/align/score.rs
+++ b/src/align/score.rs
@@ -637,6 +637,7 @@ mod tests {
         Genome {
             sequence,
             n_genome,
+            n_genome_real: n_genome,
             n_chr_real: 1,
             chr_name: vec!["chr1".to_string()],
             chr_length: vec![seq.len() as u64],
--- a/src/align/stitch.rs
+++ b/src/align/stitch.rs
@@ -386,6 +386,58 @@ pub fn cluster_seeds(
     let seed_per_window_nmax = params.seed_per_window_nmax;
     let min_seed_length = params.seed_map_min;
 
+    let n_genome_real = index.genome.n_genome_real;
+    let n_genome = index.genome.n_genome;
+    let sjdb_overhang = index.sjdb_overhang;
+    let prepared = index.prepared_junctions.as_slice();
+
+    // Expand one raw SA hit into the candidate `(real_fwd_pos, read_offset,
+    // sub_length, sub_sa_pos)` tuples cluster_seeds should consume. Hits
+    // in the real genome pass through unchanged. Hits in the Gsj
+    // flanking buffer are decoded via the sjdb table - yielding one
+    // entry for hits confined to a single flank, two entries for hits
+    // that straddle the donor/acceptor boundary (so the stitch DP can
+    // chain them through its splice branch).
+    let expand_hit = |sa_pos: u64, strand: bool, length: usize| -> Vec<(u64, usize, usize, u64)> {
+        let raw_fwd = index.sa_pos_to_forward(sa_pos, strand, length);
+        if raw_fwd < n_genome_real {
+            return vec![(raw_fwd, 0, length, sa_pos)];
+        }
+        if sjdb_overhang == 0 || prepared.is_empty() {
+            return Vec::new();
+        }
+        let mut decoded = crate::junction::sjdb_insert::decode_gsj_hit(
+            raw_fwd,
+            length,
+            n_genome_real,
+            sjdb_overhang,
+            prepared,
+        );
+        // Reverse-strand hits traverse the donor/acceptor halves in reverse
+        // read order: the leftmost forward bytes (donor flank) align to the
+        // last read bases. Swap the read offsets so each sub-seed's
+        // `read_pos = seed.read_pos + read_offset` lands at the right place
+        // in original-read coords.
+        if strand && decoded.len() == 2 {
+            let acceptor_len = decoded[1].2;
+            decoded[0].1 = acceptor_len;
+            decoded[1].1 = 0;
+        }
+        decoded
+            .into_iter()
+            .map(|(real_fwd, read_off, sub_len)| {
+                let sub_sa_pos = if strand {
+                    n_genome
+                        .saturating_sub(real_fwd)
+                        .saturating_sub(sub_len as u64)
+                } else {
+                    real_fwd
+                };
+                (real_fwd, read_off, sub_len, sub_sa_pos)
+            })
+            .collect()
+    };
+
     let anchor_set: Vec<bool> = seeds
         .iter()
         .map(|seed| {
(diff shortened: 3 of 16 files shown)


pinin4fjords:
Verified end-to-end on macOS/aarch64 against the rebuilt fix branch. Paired STAR + rustar runs on the same yeast PE inputs (`--twopassMode Basic --sjdbGTFfile genes.gtf --runRNGseed 0`, 50k reads, chrI).

`Log.final.out` splice section:

| Field | rustar pre-fix | rustar after #51 | STAR 2.7.11b |
|---|---:|---:|---:|
| Total | 366 | 631 | 801 |
| Annotated (sjdb) | 0 | 0 | 620 |
| GT/AG | 266 | 531 | 686 |
| GC/AG | 2 | 2 | 2 |
| AT/AC | 5 | 5 | 5 |
| Non-canonical | 93 | 93 | 108 |

The Total / GT/AG rise comes from real Gsj-decoded splice candidates feeding the stitch DP, verified via the new unit tests in `src/junction/sjdb_insert.rs` (all 7 pass). The decoder coverage is:

- `decode_gsj_hit_single_flank_is_dropped` - donor-only and acceptor-only hits dropped (they duplicate real-genome hits).
- `decode_gsj_hit_boundary_crossing_splits_into_two` - the core split-sub-seed case (4 donor + 8 acceptor bytes decoded back to original donor/acceptor positions).
- `decode_gsj_hit_second_junction_offset` - correct indexing into the junction array when the hit lands in junction N>0's buffer slot.
- `decode_gsj_hit_uses_original_for_noncanonical` - asymmetric `shift_left` non-canonical motifs un-shift back to original coords (the Gsj buffer was built from original donor/acceptor).
- `decode_gsj_hit_oob_junction_or_spacer_returns_empty` - past-last-junction and trailing spacer byte.
- `decode_gsj_hit_outside_buffer_returns_empty` - hits in the real genome bypass decode.
- `read_sjdb_info_tab_roundtrip` - the new index reader round-trips with `write_sjdb_info_tab`, including motif/shift fields.

`Annotated (sjdb)` stays at 0 because the annotation lookup at scoring time is the separate bug in [#45](https://github.com/scverse/rustar-aligner/pull/45); the two fixes compose to close the rest of the splice gap.

Non-canonical did not change (93 -> 93). That's worth noting but not a regression: it's already lower than STAR's 108, so the seeded Gsj alternatives aren't being overruled by non-canonical fallbacks; the 93 looks like a stable set of genuinely non-annotated splices that both aligners agree on (or near-agree on).

SJ.out.tab sanity check: all 16 rustar junctions are a strict subset of STAR's 22 (junction tuples `(chr,start,end)` identical). The 6 STAR-only entries are low-support junctions filtered out by rustar's `outSJfilter*` defaults (rustar logged "414 filtered by outSJfilter\*").

LGTM. Methodological parity with STAR's Gsj-based pass-1 seeding.

Psy-Fer:
merged in #77

PR #57, opened 2026-05-15
chore: upstream cigar

flying-sheep:
TODO: why are we rolling our own `CigarOp` instead of using `noodles::sam::alignment::record::cigar::Op`?

The only difference seems to be that that one has a `Pad` variant that we don’t have.

### DIFF
--- a/src/align/read_align.rs
+++ b/src/align/read_align.rs
@@ -355,14 +355,14 @@ pub fn align_read(
             a.genome_start,
             a.genome_end,
             a.is_reverse,
-            &a.cigar,
+            //&a.cigar,
         )
             .cmp(&(
                 b.chr_idx,
                 b.genome_start,
                 b.genome_end,
                 b.is_reverse,
-                &b.cigar,
+                //&b.cigar,
             ))
             .then_with(|| b.score.cmp(&a.score))
     });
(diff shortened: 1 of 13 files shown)


Psy-Fer:
Oh yea so funny story. 
Claude absolutely has no idea how to deal with cigar strings and messed this up so many times, that I wrote a quick version myself to get it on track. Forgot to go back and fix it up. 😅

flying-sheep:
So you’re saying that we should get rid of `CigarOp` and use the upstream version?

Psy-Fer:
I'm saying we should fix it.

flying-sheep:
Is that a yes? Otherwise I don’t know what “fix” means, as I don’t know of anything that’s broken, just duplicate code that we can eliminate.

src/align/read_align.rs
```diff
@@ -356,14 +355,14 @@ pub fn align_read(
             a.genome_start,
             a.genome_end,
             a.is_reverse,
-            &a.cigar,
+            //&a.cigar,
```
flying-sheep: This (and another `cmp` farther down) are the only casualties: why is this here? Can a vector of cigar ops be “greater” or “less” than another? (other than length)

src/align/read_align.rs
```diff
@@ -1416,18 +1412,18 @@ fn extract_junctions_from_cigar(t: &Transcript) -> Vec<(u64, u64)> {
...
-            CigarOp::RefSkip(n) => {
+            Kind::Skip => {
                 let donor = genome_pos;
-                let acceptor = genome_pos + *n as u64;
+                let acceptor = genome_pos + op.len() as u64;
                 junctions.push((donor, acceptor));
                 genome_pos = acceptor;
             }
-            CigarOp::Ins(_) | CigarOp::SoftClip(_) | CigarOp::HardClip(_) => {}
+            Kind::Insertion | Kind::SoftClip | Kind::HardClip | Kind::Pad => {}
```
flying-sheep, line 1430: I added `Pad` to all the “else” branches. Is that correct?

src/align/read_align.rs
flying-sheep: ah got it, `dedup` needs consecutive duplicates, so we first sort.

src/align/read_align.rs
flying-sheep: yes it is because `P` is only used in multiple sequence alignment which doesn’t apply here.

PR #67, opened 2026-05-17
chore: switch sam attributes to bitflags

flying-sheep:


src/io/sam.rs
```diff
@@ -104,10 +104,7 @@ impl SamWriter {
         };
         let effective_n = n_alignments.max(n_for_mapq);
         let mapq = calculate_mapq(effective_n, params.out_sam_mapq_unique);
-        let mut attrs = params.sam_attribute_set();
-        if params.out_sam_strand_field != "intronMotif" {
-            attrs.remove("XS");
-        }
```
flying-sheep, line 110: this is handle in the verification step now.

src/params/sam.rs
```diff
@@ -1,3 +1,109 @@
...
+    fn from_str(s: &str) -> Result<Self, String> {
+        Ok(match s {
+            "Standard" => Self::STANDARD,
+            "All" => Self::ALL,
+            "None" => Self::empty(),
+            "NH" => Self::NH,
+            "HI" => Self::HI,
+            "AS" => Self::AS,
+            // STAR maps NM attribute to 'nM' tag (mismatches only, not edit distance)
+            "NM" | "nM" => Self::NM,
```
flying-sheep, line 46: I didn’t add two separate flags for these as they’re treated the same.

If we want to treat them differently in the future, we can add one.

Psy-Fer:
merged in #77

PR #118, opened 2026-07-24
feat: migrate multimapper-tiebreak RNG to in-tree splitmix64

BenjaminDEMAILLE:
## Summary
- Replace the `rand` crate (`StdRng`) with STAR-rs's `splitmix64`-based deterministic generator for the two RNG consumers: `shuffle_tied_prefix` (SE/PE `--outMultimapperOrder Random` tie-break) and `pick_primary_and_mapq` (transcriptome-space primary pick).
- Per `docs-old/dev/porting-from-star-rs.md`'s RNG decision: this must land before STARsolo, whose UMI dedup depends on deterministic read order.
- Keeps the existing per-read seed (`run_rng_seed` + read-name hash, already thread-invariant) — only the underlying generator changes, matching STAR-rs's own choice of an in-tree splitmix64 over a general-purpose PRNG crate (STAR-rs's `DIVERGENCES.md` D16: STAR's own mt19937 output is itself thread-count-dependent, so no external crate's output is any more "correct" to match).
- Also migrates the test-only `packed_stream.rs` fuzz generator off `rand`, so the dependency drops from `Cargo.toml` entirely.

## Test plan
- [x] cargo build --all-targets / test / clippy -D warnings / fmt --check all green (455 tests)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -174,17 +174,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
-name = "chacha20"
-version = "0.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6f8d983286843e49675a4b7a2d174efe136dc93a18d69130dd18198a6c167601"
-dependencies = [
- "cfg-if",
- "cpufeatures",
- "rand_core",
-]
-
-[[package]]
 name = "chrono"
 version = "0.4.45"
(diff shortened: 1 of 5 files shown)


Psy-Fer:
This is worth adopting early, as it should clean up the comparisons/tests quite a lot.

We can later assess if we continue to use this method at a later date.

Psy-Fer:
If you keep all your PRs on the same base commit, I can take care of merging them all. So no need to update each one.

BenjaminDEMAILLE:
> If you keep all your PRs on the same base commit, I can take care of merging them all. So no need to update each one.

Oh fine so. will let you do the merge :)

PR #122, opened 2026-07-24
feat: --sjdbFileChrStartEnd explicit junction TSV files

BenjaminDEMAILLE:
## Summary
- Parses `chr  start  end  [strand]` TSV files (1-based intron first/last base, sister project STAR-rs's convention) into the same raw junction shape (`chr_idx, intron_start, intron_end, strand`) `--sjdbGTFfile` parsing already produces
- Unions both sources before sjdb insertion into the genome/SA (`genomeGenerate`'s `build_prep`, shared by both the in-memory and streaming index-build paths)
- Works standalone with no GTF: junction DB + sjdb files are built whenever the union is non-empty; only the transcriptome index still requires a GTF
- Unknown chromosome name in the TSV is a hard error (matches STAR's fatal exit on the same condition)

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo test` (459 passed, 4 new)
- [x] manual smoke test: `genomeGenerate` with only `--sjdbFileChrStartEnd` (no GTF) on a synthetic genome — junction round-trips correctly through `sjdbList.out.tab`, `sjdbInfo.txt` populated, no transcriptome files written

### DIFF
--- a/src/index/mod.rs
+++ b/src/index/mod.rs
@@ -273,11 +273,12 @@ impl GenomeIndex {
         // preparation + Gsj append must happen BEFORE the suffix array is
         // built, because STAR indexes the flanking-sequence buffer alongside
         // the real genome in a single SA (`sjdbBuildIndex.cpp:293`).
-        let (junction_db, transcriptome, prepared_junctions) = if let Some(ref gtf_path) =
-            params.sjdb_gtf_file
-        {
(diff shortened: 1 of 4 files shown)


BenjaminDEMAILLE:
Folded into #127 (splice-junction-DB / GTF annotation config theme) as its own commit.

PR #124, opened 2026-07-24
feat: --sjdbGTFtagExonParentGeneName/GeneType + gene_type fallback fix

BenjaminDEMAILLE:
## Summary
- Makes the GTF attribute keys used for gene name/biotype configurable via `--sjdbGTFtagExonParentGeneName` (default `gene_name`) and `--sjdbGTFtagExonParentGeneType` (default `gene_type`, `gene_biotype`)
- Multiple candidate keys are supported; when several match, the last one in the list wins (STAR's `find_attr_list` semantics)
- **Bug fix bundled in**: the previous hardcoded lookup only ever checked the `gene_biotype` attribute. GENCODE-style GTFs that use `gene_type` instead silently fell back to the literal `"MissingGeneType"`, even though STAR's own default candidate list checks `gene_type` too. Verified end-to-end: a GTF with only `gene_type "lncRNA"` now correctly writes `lncRNA` to `geneInfo.tab` instead of `MissingGeneType`.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo test` (458 passed, 3 new)
- [x] manual smoke test: `genomeGenerate` with a `gene_type`-only GTF — `geneInfo.tab` shows the correct biotype instead of `MissingGeneType`

### DIFF
--- a/src/index/io.rs
+++ b/src/index/io.rs
@@ -87,6 +87,8 @@ impl GenomeIndex {
                 &genome,
                 &params.sjdb_gtf_tag_exon_parent_transcript,
                 &params.sjdb_gtf_tag_exon_parent_gene,
+                &params.sjdb_gtf_tag_exon_parent_gene_name,
+                &params.sjdb_gtf_tag_exon_parent_gene_type,
             )?)
         } else {
             None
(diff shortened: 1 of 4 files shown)


BenjaminDEMAILLE:
Folded into #127 (splice-junction-DB / GTF annotation config theme) as its own commit.

PR #125, opened 2026-07-24
feat: --outSAMattrIHstart (HI tag start value)

BenjaminDEMAILLE:
## Summary
- Adds `--outSAMattrIHstart` (default 1, matching STAR): the start value of the `HI` SAM tag
- Threaded through every SAM/BAM record writer that emits `HI`: the SE multimapper loop, PE mate records, transcriptome-projected records, and the half-mapped-pair fallback
- The secondary FLAG bit (`0x100`) stays rank-based and is unaffected by the shift — only the printed tag value moves
- `0` gives the CellRanger convention (HI starts at 0 instead of 1)

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo test` (456 passed, 1 new)
- [x] manual smoke test: synthetic genome with a repeated 60bp region, aligned a read matching both copies — default output shows `HI:i:1`/`HI:i:2`, `--outSAMattrIHstart 0` shows `HI:i:0`/`HI:i:1`, FLAG 256 (secondary) unchanged

### DIFF
--- a/src/io/sam.rs
+++ b/src/io/sam.rs
@@ -146,6 +146,7 @@ impl SamWriter {
                 mapq,
                 max_output,    // NH = number of reported alignments
                 hit_index + 1, // 1-based
+                params.out_sam_attr_ih_start,
                 attrs,
             )?;
             maybe_insert_rg_tag(&mut record, rg_id);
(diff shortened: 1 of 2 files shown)


BenjaminDEMAILLE:
Folded into #126 (SAM output record-level tweaks theme) as its own commit.

PR #148, opened 2026-07-28
clip: CellRanger4 3' poly-A trim

BenjaminDEMAILLE:
CellRanger4 3' poly-A trim.

## What changed

`src/clip/cellranger4.rs`, ported from STAR's `ClipCR4.cpp` and `ClipMate_clipChunk.cpp`:

`poly_tail_3p` scores in from the 3' end, `+1` per `A` and `-2` per non-`A`, keeps the longest prefix of that walk still clearing a 70% density threshold, gives up once the score falls more than 27 behind, and returns nothing unless the kept score reached 20. Wired into `clip_mate_cellranger4` and reached by `--clipAdapterType CellRanger4`.

The 5' TSO clip is **not** in this PR. It needs a Smith-Waterman aligner, which per @Psy-Fer's review belongs in a separate library rather than inside rustar; `src/swalign` was removed from this branch. `--clipAdapterType CellRanger4` therefore performs the poly-A trim and logs a warning that the TSO clip is not yet applied, rather than silently doing half the job.

## Why

`--clipAdapterType CellRanger4` parsed and validated, then did nothing: `src/clip/mod.rs` declared the mode out of scope, so a user who asked for it got no clipping and the run reported success.

## One behaviour worth pointing at

The poly-A scan does not stop at the tail boundary, so A-rich sequence upstream legitimately extends the trim: `ACGT`×5 followed by 30 `A`s trims **34**, not 30. That is STAR's behaviour and reads as an off-by-one; `poly_tail_scan_does_not_stop_at_the_tail_boundary` pins it.

## Verification

- `cr4_polya_trim_matches_star` — the frozen vector
- `poly_tail_scan_does_not_stop_at_the_tail_boundary` — the behaviour above
- `test_clip_adapter_type_cellranger4` — end to end through the CLI. A 100-base read with a 40-base poly-A tail does not merely align differently untrimmed: it fails the mismatch filters and is lost entirely. Trimmed, it aligns `60M40S`. That is what the flag being inert actually cost.

Gate: 567 lib + 22 integration tests, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, MSRV 1.89 — all green

Output-neutral unless `--clipAdapterType CellRanger4` is passed.


Psy-Fer:
This is too much to be handled within rustar. This is why Star imports opal. 


I have already written a tool/library that does deterministic sw with full simd and rustar integration in rust as an opal replacement. This keeps the complexity of a whole different aligner in its own real of tests and verification. 

I mentioned this in the scverse chat as well as in the PR handling CR4 and that I would be making that tool.

James

Psy-Fer:
@BenjaminDEMAILLE you need to be the one taking responsibility and reading things, not claude.

Psy-Fer:
While integrating hyalite, I integrated what was needed from this PR too, so this will be closed when that PR lands.

PR #152, opened 2026-07-28
solo: --soloUMIfiltering MultiGeneUMI_All (closes #144)

### LINKED ISSUE #144: --soloUMIfiltering MultiGeneUMI_All is aliased to MultiGeneUMI (matches neither STAR nor the docs)
## Summary

`--soloUMIfiltering MultiGeneUMI_All` currently resolves to the same behaviour as `MultiGeneUMI`:

```rust
// src/solo/count.rs:124-125
// MultiGeneUMI_All behaves like MultiGeneUMI for the count matrix.
"MultiGeneUMI" | "MultiGeneUMI_All" => Ok(Self::MultiGeneUmi),
```

(same alias documented at `src/params/mod.rs:1090`).

That is neither what STAR 2.7.11b does nor what the option is documented to do, so rustar is currently the only one of the three behaviours that nobody asked for.

## What STAR 2.7.11b actually does

In STAR, `MultiGeneUMI_All` is **a no-op**. The option is parsed and stored, but the consumption site tests only the `MultiGeneUMI` flag, so selecting `MultiGeneUMI_All` alone leaves the multi-gene UMI filter entirely disabled. The counts you get are the unfiltered counts.

## What it is documented to do

STAR's documentation describes the two as:

- `MultiGeneUMI`: remove the lower-count gene's reads when a UMI is assigned to more than one gene.
- `MultiGeneUMI_All`: when a UMI is assigned to more than one gene, remove it from **all** of those genes, not just the losers.

So `MultiGeneUMI_All` is strictly more aggressive than `MultiGeneUMI`. Aliasing it to `MultiGeneUMI` silently under-filters; STAR's own no-op silently does not filter at all.

## Three options

1. **Reproduce STAR's no-op.** Maximum byte-fidelity: a run with `--soloUMIfiltering MultiGeneUMI_All` matches STAR exactly. Also means the flag does nothing, which will surprise anyone who reads the STAR docs.
2. **Implement the documented behaviour** (split the enum variant, drop cross-gene UMIs from every gene they touch). Diverges from STAR on this one flag, with a documented entry and a locking test.
3. **Status quo** (alias to `MultiGeneUMI`). Matches neither.

## Resolution

Option 2, implemented in #152:

- a separate `UmiFiltering::MultiGeneUmiAll` variant rather than an alias;
- an entry in a new `docs-old/dev/divergences.md` recording the deviation and the reason;
- `test_solo_multigene_umi_all_drops_cross_gene_umis` asserting the correct result, never STAR's no-op.

Only the `MultiGeneUMI_All` path changes; the default (`-`) and the other modes produce identical output. If option 1 is preferred instead, the test asserts the behaviour either way and inverting it is a one-line change.

BenjaminDEMAILLE:
`--soloUMIfiltering MultiGeneUMI_All` does what it is documented to do. Closes #144.

## What changed

`UmiFiltering::MultiGeneUmiAll` becomes its own variant instead of an alias, and removes a UMI seen in more than one gene from all of them.

| | behaviour |
|---|---|
| STAR 2.7.11b | no-op — parsed and stored, but the consumption site tests only the `MultiGeneUMI` flag, so selecting it leaves the filter off |
| STAR's documentation | remove a UMI seen in more than one gene from **all** of them |
| this codebase, before | same as `MultiGeneUMI`: keep the deepest-read gene |

## Why

Reproducing the no-op ships a flag that silently does nothing to anyone who read STAR's documentation. #144 records the choice, and it was raised there before any code changed, because "be faithful to STAR" and "do what the flag says" genuinely point in opposite directions here.

## Divergence

`DIVERGENCE.md` §1.2, with the STAR source checked (`SoloFeature_collapseUMIall.cpp`, `ParametersSolo.cpp`). Needs maintainer sign-off per CONTRIBUTING. Inverting it is a one-line change, since the test asserts the behaviour rather than the direction.

## Verification

- `multigene_umi_all_drops_the_umi_from_every_gene` — a UMI seen in two genes disappears from both, where `MultiGeneUMI` keeps the deeper one
- `multigene_umi_all_parses_to_its_own_variant` — it is no longer an alias
- single-gene UMIs are asserted untouched across every filtering mode, which is the property most at risk from a filter written in a hurry

Gate: 562 lib + 26 integration tests, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, MSRV 1.89 — all green.

This is a counting change on the `MultiGeneUMI_All` path only. The default (`-`) and the other modes produce identical output, so no existing run changes.

**Not run: `test/solo_diff_docker.sh`.** Since the default is unaffected, the harness would compare an unchanged path; the changed path has no STAR behaviour to compare against, because in STAR it does nothing. Sign-off on the divergence is what this needs rather than an oracle run.

This PR was previously bundled with four other solo themes; those are now #156 (EmptyDrops_CR statistics), #157 (CellReads.stats), #158 (Transcript3p) and #159 (soloCellFiltering), per the one-theme rule in CONTRIBUTING.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,12 @@ Sections commonly used: Features, Bug fixes, Other changes.
 
 ### Bug fixes
 
+- `--soloUMIfiltering MultiGeneUMI_All` was aliased to `MultiGeneUMI`,
+  which is neither STAR's behaviour nor the documented one: in STAR
+  2.7.11b the variant is a no-op. It now removes a UMI seen in two or
+  more genes from **all** of them, the behaviour the option name
+  describes. Recorded in `DIVERGENCE.md` (closes #144).
+
 - **STARsolo `Gene` assignment now requires exon concordance**, matching
   STARsolo: a read counts toward a gene only when every aligned block
   lies within the gene's exons, rather than merely overlapping one. This
--- a/DIVERGENCE.md
+++ b/DIVERGENCE.md
@@ -42,18 +42,6 @@ This is the reason faithfulness is reported **tie-adjusted**. On the 10k yeast b
 
 **Source.** `src/solo/count.rs` (`UmiFiltering::MultiGeneUmiAll`, `filter_multi_gene_umi`), locked by `test_solo_multigene_umi_all_drops_cross_gene_umis`. STAR: `SoloFeature_collapseUMIall.cpp`, `ParametersSolo.cpp`.
 
-### 1.3 Homopolymer UMIs other than poly-A are rejected
-
-**What STAR does.** The UMI validity check rejects a homopolymer by comparing the packed UMI against a precomputed all-same-base value, but the loop that builds those values runs over `umiL`, which is zero at that point for `CB_UMI_Simple`. Only the poly-A case (packed value zero) is caught; poly-C, poly-G and poly-T pass through as valid UMIs.
-
-**What rustar-aligner does.** Rejects every homopolymer UMI.
-
-**Why.** A homopolymer UMI is a sequencing artefact whatever base it repeats; letting three of the four through is not a rule, it is the consequence of reading an uninitialised length.
-
-**Impact.** Removes a small number of artefact UMIs from the counts that STAR keeps.
-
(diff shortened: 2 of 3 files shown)


Psy-Fer:
Closed by #212

PR #156, opened 2026-07-29
solo: EmptyDrops_CR on CellRanger's actual statistics (SGT ambient profile, libc++ sampler)

BenjaminDEMAILLE:
`--soloCellFilter EmptyDrops_CR` computes what CellRanger computes: a Simple-Good-Turing ambient profile, sampled with libc++'s generator.

## What changed

**The ambient profile is smoothed with Simple Good-Turing** (`src/solo/sgt.rs`), the method CellRanger uses and STAR vendors as Elworthy's `sgt.h`. The ambient counts come from a small sample of empty droplets, so a gene seen twice there is not twice as likely as one seen once, and a gene seen zero times is not impossible — it is one the sample was too small to show. SGT reserves mass for the unseen from the singleton rate and smooths the rest along a fitted log-log line.

What this replaces had the right shape and the wrong numbers: it reserved unseen mass the same way, then distributed the remainder in proportion to raw counts with no smoothing at all.

**The Monte-Carlo null is drawn with libc++'s `std::mt19937` and `std::discrete_distribution`** (`src/solo/libcxx_rng.rs`), seeded `19760110 * (isim + 1)` per simulation as STAR seeds it. The standard fixes mt19937's output but not how `generate_canonical` consumes it, nor how `discrete_distribution` maps a uniform draw onto categories, so reproducing STAR's numbers needs *libc++'s* algorithm specifically, not a correct categorical sampler.

The module is wired into the EmptyDrops path in this same PR — it is not a module landed ahead of its use.

## Why

`solo::count` sampled with a `SplitMix64` stream under a comment calling it "WeightedIndex-equivalent; empirically byte-identical EmptyDrops cell calls". That claim cannot hold in general: two unrelated generators cannot agree over an arbitrary number of draws. It was true of whatever cases were checked and unknown everywhere else.

## Divergence

`DIVERGENCE.md` §1.2. STAR leaves `PZero` uninitialised when the ambient spectrum has fewer than five distinct frequencies and `analyse()` bails, so it reads whatever the stack held. Here it is zero from construction. Degenerate inputs only, but on those an uninitialised read can place arbitrary mass on unseen genes.

## Verification

Every expected value in the RNG tests was produced by compiling a C++ program against real libc++ (`clang++ -stdlib=libc++`) and printing the results, not derived from reading its source. `tests/libcxx_oracle.cpp` is in-tree so they can be regenerated. `generate_canonical` is asserted on **bit patterns** rather than decimals: a difference in the last place changes which category a sample lands in.

SGT tests cover a well-formed spectrum summing to one, monotonicity in the observation count, the absent-count case, the no-singletons case, and the D17 guard.

Gate: 570 lib + 26 integration tests, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, MSRV 1.89 — all green.

`test/solo_diff_docker.sh` (one run, Linux container with STAR 2.7.11b) passes:

```
PASS: rustar-aligner matrix matches real STARsolo exactly.
PASS: Gene raw matrix byte-identical to STARsolo (3 entries)
PASS: GeneFull raw matrix byte-identical to STARsolo (3 entries)
```

**Read that for what it is.** The harness exercises the raw matrix, which these changes do not touch, so it confirms nothing was broken on the way past — not that the cell calls are right. Both changes move `filtered/` membership by construction, and the harness has no EmptyDrops fixture to compare against. Validating the cell calls themselves needs a dataset with a known CellRanger `filtered/` result, which the harness does not currently carry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,16 @@ Sections commonly used: Features, Bug fixes, Other changes.
   glibc's malloc and per-thread heaps that return whole segments to
   the OS when abandoned, so allocator cache size stays bounded.
 
+- **`--soloCellFilter EmptyDrops_CR` now uses CellRanger's actual
+  statistics.** The ambient profile is smoothed with Simple Good-Turing,
+  as CellRanger and STAR do, instead of an approximation that reserved
+  unseen mass from the singleton rate and spread the remainder in
+  proportion to raw counts. The Monte-Carlo null is drawn with libc++'s
+  `std::mt19937` and `std::discrete_distribution`, seeded
+  `19760110 * (isim + 1)` per simulation as STAR seeds it, replacing a
+  SplitMix64 stream that could not agree with STAR's over an arbitrary
+  number of draws. Cell calls move as a result.
+
 ### Bug fixes
 
 - **STARsolo `Gene` assignment now requires exon concordance**, matching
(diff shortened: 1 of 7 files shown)


Psy-Fer:
Closed by #212

PR #157, opened 2026-07-29
solo: --soloCellReadStats CB writes CellReads.stats

BenjaminDEMAILLE:
`--soloCellReadStats CB` writes `Solo.out/<feature>/CellReads.stats`.

## What changed

One row per cell barcode with the fifteen counters STARsolo reports: how the barcode matched (`cbMatch`, `cbPerfect`, `cbMMunique`, `cbMMmultiple`), whether the read mapped to one genomic locus or several (`genomeU`/`genomeM`), whether it landed on a feature (`featureU`/`featureM`), where in the gene and on which strand (`exonic`, `intronic`, `exonicAS`, `intronicAS`), whether it was mitochondrial (`mito`), and whether it reached the matrix (`countedU`/`countedM`). Then `nUMIunique` and `nGenesUnique`, taken from the raw matrix so they agree with it by construction.

`--genomeChrSetMitochondrial` (new) names the chromosomes behind the `mito` column. Without it the column is zero throughout, which is honest: no chromosome was declared mitochondrial.

Reads whose barcode never resolved are summed into a single `CBnotInPasslist` row rather than dropped. That row is the reason the file is useful: it separates "these cells look thin" from "most of the input never reached a cell at all".

The region columns split by strand — an antisense read counts under `exonicAS` or `intronicAS`, never under `exonic` or `intronic`.

## Divergence

`DIVERGENCE.md` §3.2. STAR emits these rows by walking a libc++ `unordered_map`, which at these map sizes is the reverse of each barcode's first appearance. That is reproduced, including across threads: the accumulator merges in read order, so a threaded run writes the same file as a serial one. It stops being reproducible past the point where libc++ rehashes, since the order then depends on the bucket count. The values never differ, only which line they sit on.

## Verification

Six unit tests in `src/solo/cell_reads.rs`:

- `each_row_fills_the_header` — twenty columns, every row complete
- `flags_accumulate_per_read` — `cbMatch` counts every read; each flag counts only its own
- `unresolved_reads_land_in_the_passlist_miss_row`
- `rows_are_emitted_in_reverse_first_appearance_order` — the divergence above
- `umi_and_gene_totals_come_from_the_matrix` — a cell absent from the matrix prints zeros rather than inheriting a read count

Reads are folded in under a mutex held by `SoloContext`, so the row order is the order reads were processed in regardless of `--runThreadN`.

Gate: 565 lib + 26 integration tests, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, MSRV 1.89 — all green.

Output-neutral: the file is written only when `--soloCellReadStats CB` asks for it, and nothing else changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,27 @@ Sections commonly used: Features, Bug fixes, Other changes.
   glibc's malloc and per-thread heaps that return whole segments to
   the OS when abandoned, so allocator cache size stays bounded.
 
+- **`--soloCellFilter EmptyDrops_CR` now uses CellRanger's actual
+  statistics.** The ambient profile is smoothed with Simple Good-Turing,
+  as CellRanger and STAR do, instead of an approximation that reserved
+  unseen mass from the singleton rate and spread the remainder in
+  proportion to raw counts. The Monte-Carlo null is drawn with libc++'s
+  `std::mt19937` and `std::discrete_distribution`, seeded
+  `19760110 * (isim + 1)` per simulation as STAR seeds it, replacing a
+  SplitMix64 stream that could not agree with STAR's over an arbitrary
+  number of draws. Cell calls move as a result.
+
+- **`--soloCellReadStats CB`** writes `Solo.out/<feature>/CellReads.stats`:
+  one row per cell barcode with fifteen counters describing what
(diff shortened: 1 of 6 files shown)


Psy-Fer:
Landed in #200

Issue #168, opened 2026-07-30
perf: 1000 heap allocations per read in the align path (and how to fix it)


BenjaminDEMAILLE:
## What I measured

Wrapping the global allocator in a counter and running 200 000 real reads (yeast, `ERR12389696`) at `--runThreadN 8`:

```
ALLOCSTATS count=200093070 bytes=401267330431
```

**1 000 heap allocations per read, and 2 MB allocated per read.**

For context, at 16 threads on 2M reads a sampled profile attributes roughly 8% of on-CPU time to mimalloc itself (`mi_bchunk_try_find_and_clearNX`, `mi_free`, `mi_malloc_aligned`, `mi_bitmap_clearN`) and another ~3.5% to `_platform_memmove`. So this is the second-largest cost in the aligner after `find_best_junction_position`, and unlike that one it is not intrinsic work — it is bookkeeping.

## Where it comes from

`stitch_recurse` clones a `WorkingTranscript` at each include/exclude branch (`src/align/stitch.rs:1195`, `:1320`, `:2403`), and `MAX_RECURSION` is 100 000 per cluster. Each clone copies four `Vec`s:

```rust
pub(crate) struct WorkingTranscript {
    pub(crate) exons: Vec<ExonBlock>,
    pub(crate) junction_motifs: Vec<SpliceMotif>,
    pub(crate) junction_annotated: Vec<bool>,
    pub(crate) junction_shifts: Vec<(u32, u32)>,
    ...
}
```

The three junction vectors are empty on an unspliced read, and an empty `Vec` does not allocate, so the steady-state cost is `exons` plus whatever the spliced reads add. `cluster_seeds` contributes its own per-read set: `anchor_set`, `anchor_indices`, `windows`, the `win_bin` map, a `Vec` per window, and `diag_ranges` with a `Vec` per diagonal.

## What I am *not* doing without asking

The obvious fix is inline storage for those four vectors, which in practice means `smallvec` with inline capacities around 8/4/4/4. `CONTRIBUTING.md` says a new dependency has to be raised before the PR, so I am raising it rather than opening one.

Three ways I can see, and I have no strong preference between the first two:

1. **`smallvec`.** Widely used, no build-script or C, MSRV-friendly. One more dependency on a project that publishes to crates.io and builds on five platforms.
2. **A small in-tree inline vector.** No dependency, but it is `unsafe`-adjacent code to review and maintain for a type this project would only use in one place.
[...]

PR #170, opened 2026-07-30
refactor: drop Transcript::read_seq, a public field nothing reads

BenjaminDEMAILLE:
`Transcript::read_seq` is filled with a full copy of the read at every finalised alignment, and nothing in the crate reads it.

## How I know nothing reads it

Not from grepping. I deleted the field and asked the compiler. All 97 resulting errors are struct literals or assignments:

```
error[E0560]: struct `Transcript` has no field named `read_seq`
  --> src/align/stitch.rs:2158:9    read_seq: read_seq.to_vec(),
error[E0609]: no field `read_seq` on type `Transcript`
  --> src/align/stitch.rs:2698:28   transcript.read_seq = read_seq.to_vec();
  --> src/align/read_align.rs:892   t1.read_seq = mate1_seq.to_vec();
  ...
```

Not one read site, in `src` or in tests. The three places that looked like readers on a grep are something else: `io/sam.rs:265` uses a local parameter of the same name, and `chimeric/output.rs:148,157` read `ChimericAlignment::read_seq`, a different struct filled from its own parameter in `chimeric/detect.rs`.

## What it costs

Measured on 200 000 real reads (`ERR12389696`, yeast) at 8 threads, with the global allocator wrapped in a counter:

| | allocations | bytes |
|---|---|---|
| before | 200 093 070 | 401 267 330 431 |
| after | 184 966 576 | 397 400 813 793 |

**15.1 million allocations removed, 7.6% of the total, and 3.87 GB of copying.** About 50 transcripts are finalised per read, each copying the whole read.

## Wall clock: no change, and that is the interesting part

Six interleaved rounds, 2M reads, `--runThreadN 16`, 87-92% CPU idle:

| | median |
|---|---|
| before | 20.90 s |
| after | 20.65 s |

Direction mixed, inside the run-to-run spread. One round had a clear outlier which I have not excluded from the medians.

So removing 7.6% of all allocations moves nothing measurable. mimalloc is fast, and it is already tuned in `src/main.rs`. **That calibrates #168 downward**: I estimated an ~11% ceiling for allocation work from the profile's mimalloc share, and this says that estimate is too generous. I would not spend a new dependency on the rest of it without a workload that shows the cost, and I have updated #168 accordingly.

The reason to take this change is that the field is dead weight, which `CONTRIBUTING.md` rules out on its own terms. The allocations are a bonus, not the argument.

## One thing needing your call

`Transcript::read_seq` is `pub`, so this is an API removal. Nothing outside the crate can be relying on its *contents*, since it is only ever written, but the field disappearing is still a breaking change for anyone naming it. Say the word if you would rather keep it as a deprecated always-empty field for a release.

## Verification

Output-neutral: SAM byte-identical on 200 000 real reads against `origin/main`, whole file including header.

Gate: 560 lib + 26 integration tests, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, all green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,3 +131,9 @@ Sections commonly used: Features, Bug fixes, Other changes.
   needs random access to the SA in RAM.
 
 Initial release of Rust rewrite of STAR.
+### Other changes
+
+- Removed `Transcript::read_seq`, a public field that was filled with a
+  copy of the read at every finalised alignment and never read. **API
+  removal.** Output is unchanged.
+
--- a/src/align/pe_overlap.rs
+++ b/src/align/pe_overlap.rs
@@ -364,7 +364,6 @@ pub fn convert_merged_transcript_to_pe(
             n_junction,
             junction_motifs: out_junctions[i].iter().map(|(m, _)| *m).collect(),
             junction_annotated: out_junctions[i].iter().map(|(_, a)| *a).collect(),
-            read_seq: mate_seqs[i].to_vec(),
         });
     }
 
@@ -479,7 +478,6 @@ mod tests {
             n_junction: 0,
             junction_motifs: vec![],
             junction_annotated: vec![],
-            read_seq: merge.merged.clone(),
         };
 
         let scorer = AlignmentScorer::from_params_minimal();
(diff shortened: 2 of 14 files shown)


Psy-Fer:
landed in #200

PR #171, opened 2026-07-30
test: A/B bench harness that drops rounds it could not measure honestly

BenjaminDEMAILLE:
`test/bench_ab.sh`: interleaved A/B of two binaries, which drops the rounds it could not measure honestly instead of averaging over them.

## Why

Four traps, each of which I fell into at least once while measuring this aligner over one session. Three are now enforced by the script; the fourth is documented in its header because no script can enforce it.

**Idle was checked once, at the start.** A previous harness sampled CPU idle before the first run and never again. Three separate measurements then drifted below the threshold mid-series and it kept going, leaving the contaminated rounds to be spotted by eye. This samples idle before *and* after every run, drops the round if any of the four samples falls short, and prints the count:

```
round1  binA=3.33s binB=3.42s  idle>=84.92%
round3  binA=3.25s binB=3.22s  DROPPED (idle fell to 71.4%)

kept 4 of 6 rounds (2 dropped)
```

A median over four surviving rounds now cannot be read as a median over six. No silent truncation.

**Load average is the wrong signal.** It is an exponential average over minutes, so it stays high long after the offending job has gone. An earlier guard refused to measure at 2.24 on a machine whose sixteen cores were all free. The check is on CPU idle percentage, which answers the question actually being asked.

**A median can be smaller than the noise.** The script prints the spread within each side next to the difference between the medians, and when the difference is smaller it says so:

```
median  binA=3.305s  binB=3.25s  delta=-1.7%
spread  binA=0.08s   binB=0.20s

the difference between the medians is smaller than the spread within a side.
that is not a result: report it as unmeasured rather than as a speedup.
```

That message is not hypothetical. Two changes I measured this week looked like wins on medians alone and were inside the spread; both are now described as unmeasured in their PRs (#169, #170).

**Both sides must see identical argv.** `@PG` `CL:` records the command line verbatim, so running `./old` against `./new`, or writing to differently-named output directories, makes the BAM differ in the header and in its compressed size. That cost me two false "output is not neutral" alarms. Each side is copied to `rustar-aligner` inside its own directory and run with `--outFileNamePrefix ./`.

**Timing a total hides the part that changed.** Only in the header, since it is a judgement the script cannot make: BAM writing is 1-4% of a yeast run, so a total dominated by alignment cannot resolve a change to the writer at any reachable scale. Run one configuration with `--outSAMtype None` alongside and read the difference. #164 was closed because that difference, once isolated, showed the change was a regression rather than the improvement its description claimed.

## Usage

```bash
test/bench_ab.sh <binA> <binB> <genomeDir> <reads.fastq> [threads] [rounds]

BENCH_MIN_IDLE=88   percent CPU idle a round must hold to count
BENCH_MODE="None"   --outSAMtype words (quote the two-word forms)
BENCH_ARGS=""       extra flags passed to both sides
```

It exits non-zero if every round was dropped, rather than printing a median over nothing.

## Scope

Test tooling only: no source change, no dependency, nothing user-facing, so no `CHANGELOG` entry. The numbers in #167 and #170 were produced with the same method; this makes the method reproducible rather than living in my shell history.

Happy to add a pointer to it from `CONTRIBUTING.md`'s verification section if you want one there, but I did not want to edit that file uninvited.

Related: #168 has the allocation measurements taken with this method, including the result that removing 7.6% of the allocations moved nothing.

### DIFF
--- a/test/bench_ab.sh
+++ b/test/bench_ab.sh
@@ -1,0 +1,148 @@
+#!/usr/bin/env bash
+# Interleaved A/B of two rustar-aligner binaries.
+#
+#   test/bench_ab.sh <binA> <binB> <genomeDir> <reads.fastq> [threads] [rounds]
+#
+# Prints one line per round and a median over the rounds that were measured on
+# a quiet machine, plus a count of the ones that were not. Read the whole
+# output: a median over four surviving rounds out of six means something
+# different from a median over six.
+#
+#     BENCH_MIN_IDLE=88   percent CPU idle a round must hold to count
+#     BENCH_MODE="None"   --outSAMtype words (quote the two-word forms)
+#     BENCH_ARGS=""       extra flags passed to both sides
+#
+# ---------------------------------------------------------------------------
+# Why this script exists, and what it refuses to do
+#
+# Three things went wrong often enough while measuring this aligner that they
+# are worth encoding rather than remembering.
+#
+# 1. A busy machine does not add noise, it inverts results. An unrelated job
+#    saturating the cores lands on whichever side happens to run under it, and
+#    nothing in the timings shows that it happened: the series just drifts. So
+#    idle is sampled *before and after every single run*, not once at the
+#    start, and a round where either sample falls below the threshold is
+#    dropped and counted rather than quietly folded into the median.
+#
+#    The check is on CPU idle, not load average. Load average is an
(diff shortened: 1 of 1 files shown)


Psy-Fer:
Landed in #200

PR #174, opened 2026-07-31
feat(solo): --soloOutRawBarcodes Observed — a CellRanger-shaped raw matrix (stacked on #173)

BenjaminDEMAILLE:
**Stacked on #173** — contains its commits. Merge #173 first; this rebases to a single commit afterwards.

Comparing our raw matrix against a real `cellranger count` run gave **0 identical entries out of 27 396**. Not a rounding difference: zero overlap, because "raw" means two different things.

## The finding

I ran CellRanger 10.0.0 on the fixture from #172 (`cellranger mkref` on the yeast reference, then `cellranger count`), and its raw matrix has **200 barcode columns** — the ones observed. STARsolo's has **3 686 400** — the whole v3 whitelist, nearly all zeros. Their `barcodes.tsv` is a few kilobytes; ours is 62 MB.

Two normalisations were needed before the matrices could be compared at all, and the other one is not in this PR: CellRanger appends a `-1` GEM-well suffix to every barcode. That is a separate change and I would rather not bundle it.

## What this adds

`--soloOutRawBarcodes Observed` narrows the raw matrix to barcodes carrying a count. Default `Whitelist` is what STARsolo writes, so nothing changes for anyone not asking.

| | barcodes | `barcodes.tsv` |
|---|---|---|
| `Whitelist` (default) | 3 686 400 | 62 668 800 bytes |
| `Observed` | 200 | 3 400 bytes |

**Counts are identical either way**: 13 937 entries, 15 414 counts on both sides, checked entry by entry. Only the columns present differ.

`finalize_matrix` already accepted a column remap for the filtered matrix, so this reuses that path instead of adding a second one. The observed set is read back from the streamed body — one extra pass, and only when the flag is on.

## Non-STAR flag, needs sign-off

`--soloOutRawBarcodes` does not exist in STAR. `CONTRIBUTING.md` requires that to be stated rather than presented as parity, so it is in `DIVERGENCE.md` §3.2 with the rationale, and it is off by default.

I am not attached to the name. If you would rather have one umbrella flag for CellRanger-shaped output — this, the `-1` suffix, the directory names — say so and I will fold them together instead.

## Where this sits

**With #165 applied**, and with #173, the count matrix is within **0.13% of real CellRanger** on this fixture (15 091 against 15 111; STAR itself is at 15 124, +0.09% the other way). #165 is a precondition, not a footnote: without it the same stack sits at 15 439, +2.17%. It should merge first. Numbers and method in #172, and the full table is in the comment thread below.

This PR changes no counts at all. It changes which columns are written, which is what stops the two outputs being comparable in the first place.

## Verification

- counts identical between the two modes, entry by entry, on the 20 000-read fixture
- gate: 561 lib + 26 integration tests, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, all green

Not covered by a unit test yet: the flag is exercised end to end above but the fixture that proves it lives outside the repo, alongside the replacement for the three-entry differential (item 4 of #172). I would rather land that fixture once and have it cover this, #173 and the CB-match work together than write a weaker synthetic test three times.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,11 @@ Sections commonly used: Features, Bug fixes, Other changes.
 
 - Read names are cut at `--readNameSeparator` (default `/`), as STAR does. A
   read named `foo/1` was previously emitted as `foo/1` where STAR emits `foo`.
+- `--soloOutRawBarcodes Observed` writes the raw matrix with one column
+  per *observed* barcode instead of one per whitelist barcode, matching
+  what CellRanger's `raw_feature_bc_matrix` contains. Counts are
+  unchanged; on a 200-cell run `barcodes.tsv` goes from 62 MB to 3.4 kB.
+  **Not a STAR parameter**; default `Whitelist` keeps STARsolo behaviour.
 
 - **STARsolo single-cell quantification (`--soloType`)** — the 10x
   Chromium / plate-based count-matrix pipeline, ported from STAR and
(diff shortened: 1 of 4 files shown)


BenjaminDEMAILLE:

|---|---|---|---|

### What survives

### What does not survive

### How it happened

BenjaminDEMAILLE:

|---|---|---|---|

PR #175, opened 2026-07-31
fix(solo): MultiGeneUMI_CR decides ownership on corrected UMIs (stacked on #174)

BenjaminDEMAILLE:
**Stacked on #174, which is stacked on #173.** Merge in that order; this rebases to one commit afterwards.

Finishes the `MultiGeneUMI_CR` rule. #173 fixed the tie; this adds the condition I explicitly deferred there, and it needed the restructure I said it would.

## What STAR actually does

Two things, in this order (`SoloFeature_collapseUMIall.cpp:134-148`, then `:203-235`):

**UMIs are corrected within each gene first**, and the per-gene read totals are recorded twice — once under the raw UMI (`umiGeneMapCount0`), once under the corrected one (`umiGeneMapCount`).

**Then ownership is decided on the corrected map**, subject to two conditions:

1. one gene holds a **strictly** higher read count than every other — a tie counts for nobody. That is #173.
2. and that winner is **not beaten in the uncorrected map at the same key**:

```cpp
for (const auto &ig : umiGeneMapCount0[iu.first]) {
    if (ig.second>umiGeneMapCount0[iu.first][maxg]) { maxg=-1; break; };
};
```

The second condition exists because correction moves reads between UMIs. A gene can end up winning only because correction folded a neighbouring UMI onto it, and STAR refuses that win instead of counting it.

## Why this could not be a smaller change

The generic path here filters multi-gene UMIs first and corrects afterwards. That order cannot express either condition: by the time correction runs, ownership is already decided. So `MultiGeneUMI_CR` now takes its own path through the per-cell loop, which mirrors STAR — the flag is only valid with `--soloUMIdedup 1MM_CR`, so no other combination is affected.

`cellranger_1mm_map` exposes the mapping `cellranger_1mm` was already computing and discarding.

## Measured against CellRanger 10.0.0

Real `cellranger count`, not a proxy: `cellranger mkref` on the yeast reference, then `cellranger count` on the fixture from #172. With #165 and #173 also applied:

| | identical entries | CellRanger | rustar | gap |
|---|---|---|---|---|
| #165 + #173 | 13 651 / 13 709 | 15 111 | 15 091 | −20 |
| **plus this** | **13 676 / 13 709 (99.76%)** | 15 111 | **15 116** | **+5 (+0.03%)** |

Entries CellRanger has and we do not: **29 → 7**.

**#165 is a precondition for these numbers** and should merge first; without it the same stack sits at 15 439, +2.17%. With it, STAR 2.7.11b on the same fixture with the same flags counts 15 124, so it sits +13 from CellRanger while this sits +5. All three agree to within a fraction of a percent. The full table is in the comment thread below.

## Verification

- `multi_gene_umi_cr_rejects_a_winner_that_only_wins_after_correction` — two UMIs one substitution apart, arranged so a gene wins after correction and loses before it
- `multi_gene_umi_cr_drops_a_tie_entirely` and `multi_gene_umi_cr_keeps_top_gene` from #173 still pass unchanged

Gate: 562 lib + 26 integration tests, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, all green.

## What is still different

7 entries CellRanger has that we do not, 19 the other way, of 13 709. I have not chased them and will not guess at a cause here.

The honest limit of this method: anything left is small enough that telling a genuine algorithmic difference from a fixture artefact needs a second, differently-shaped dataset. That belongs with the fixture work in #172 rather than with this rule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,12 @@ Sections commonly used: Features, Bug fixes, Other changes.
 
 ### Bug fixes
 
+- `--soloUMIfiltering MultiGeneUMI_CR` kept every gene tied at the
+  highest read count; CellRanger gives a tied UMI to no gene at all.
+  Since one read per gene is the ordinary shape of a multi-gene UMI, the
+  flag removed nothing in practice. On a 20k-read 10x fixture the count
+  matrix moves from 16 465 to 15 414 against STAR's 15 423.
+
 - **STARsolo `Gene` assignment now requires exon concordance**, matching
   STARsolo: a read counts toward a gene only when every aligned block
   lies within the gene's exons, rather than merely overlapping one. This
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ Sections commonly used: Features, Bug fixes, Other changes.
 
 ### Features
 
+- `--soloOutRawBarcodes Observed` writes the raw matrix with one column
+  per *observed* barcode instead of one per whitelist barcode, matching
+  what CellRanger's `raw_feature_bc_matrix` contains. Counts are
+  unchanged; on a 200-cell run `barcodes.tsv` goes from 62 MB to 3.4 kB.
+  **Not a STAR parameter**; default `Whitelist` keeps STARsolo behaviour.
+
 - **STARsolo single-cell quantification (`--soloType`)** — the 10x
   Chromium / plate-based count-matrix pipeline, ported from STAR and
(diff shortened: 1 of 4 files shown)


BenjaminDEMAILLE:

|---|---|---|---|

### What survives

### What does not survive

### How it happened

BenjaminDEMAILLE:

|---|---|---|---|

Psy-Fer:
Closed by #212

PR #176, opened 2026-07-31
feat(solo): CellRanger behaviour by default on 10x geometry (changes default output, stacked on #175)

BenjaminDEMAILLE:
**Stacked on #175 → #174 → #173.** Merge in that order.

**This changes default output behaviour on 10x runs.** I am putting that in the first line rather than the last, because it is the thing to argue about and everything else in the PR is arithmetic.

## The problem

A user aligns 10x data, compares the count matrix against CellRanger, and gets a successful run with different numbers. Nothing in the output points at the five flags that explain the difference; they are in STAR's `docs/STARsolo.md` and nowhere the run can see.

Measured against **CellRanger 10.0.0** on the 20 000-read fixture from #172, those five flags are the whole gap:

| | count matrix | gap |
|---|---|---|
| without them | 16 465 | +8.9% |
| with them (this stack, over #165) | 15 116 | **+0.03%** |

## What this does

On geometry that is unambiguously 10x — `CB_UMI_Simple`, a whitelist, a 16-base cell barcode, and a 10- or 12-base UMI — these default to their CellRanger values:

```
--clipAdapterType      CellRanger4
--outFilterScoreMin    30
--soloCBmatchWLtype    1MM_multi_Nbase_pseudocounts
--soloUMIfiltering     MultiGeneUMI_CR
--soloUMIdedup         1MM_CR
```

and the run says so:

```
INFO 10x geometry detected (CB 16 + UMI 12 with a whitelist): defaulting to
CellRanger behaviour [CellRanger4, 30, 1MM_multi_Nbase_pseudocounts,
MultiGeneUMI_CR, 1MM_CR]. Pass the flags explicitly to override; this differs
from STARsolo's defaults.
```

Verified end to end: with no CellRanger flags on the command line, the output is byte-for-byte what passing all five produces.

## The three things that make it defensible, and the one that does not

**It is escapable.** A flag named on the command line always wins, including when the value asked for is STARsolo's own default. `value_source` distinguishes an explicit flag from a default, so `--soloUMIdedup 1MM_All` gets `1MM_All`. A divergence you cannot turn off would be a different and much worse thing than a divergent default.

**It is narrow.** 16 bp CB with a 10 or 12 bp UMI *and* a whitelist is the geometry of every 10x 3'/5' gene expression chemistry and nothing else in common use. Non-10x geometry is untouched, and there is a test for that.

**It is loud.** Every run it touches announces the substitution and the geometry that triggered it.

**What it is not: STAR-faithful.** It is a deliberate change of default output, recorded in `DIVERGENCE.md` §1.3 as the largest entry in that file. `CONTRIBUTING.md` says this needs sign-off, and I am asking for it rather than assuming it.

If you would rather have the detection without the action — same geometry check, same message, no flags changed — that is a two-line change to this PR and I will make it on request. It was my own recommendation; the project owner asked for the default instead, and I would rather implement what was asked and let you rule on it with the code in front of you than argue it in the abstract.

## Verification

- `ten_x_geometry_defaults_to_cellranger_behaviour` — all five, from a bare 10x command line
- `an_explicit_flag_beats_the_10x_default` — including asking for STARsolo's own default
- `non_10x_geometry_keeps_starsolo_defaults` — a 12-base barcode changes nothing
- `ten_x_lengths_without_a_whitelist_keep_starsolo_defaults` — no whitelist, no override

Gate: 566 lib + 26 integration tests, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, all green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,6 @@ Sections commonly used: Features, Bug fixes, Other changes.
 
 ### Features
 
-<<<<<<< HEAD
-<<<<<<< HEAD
 - **CLI and output parity: SAM/SJ/read-input knobs and the STAR limit
   surface** — 30 further STAR 2.7.11b parameters. (`--outSAMorder` came from #145.)
 
@@ -46,21 +44,17 @@ Sections commonly used: Features, Bug fixes, Other changes.
 
 - Read names are cut at `--readNameSeparator` (default `/`), as STAR does. A
   read named `foo/1` was previously emitted as `foo/1` where STAR emits `foo`.
-=======
-=======
 - On 10x geometry (`CB_UMI_Simple`, a whitelist, 16 bp CB, 10 or 12 bp
   UMI), the five CellRanger-matching flags now **default** to their
   CellRanger values. Any flag named on the command line wins, and the
   substitution is logged. **This changes default output on 10x runs** and
   diverges from STARsolo; see `DIVERGENCE.md` §1.3.
 
->>>>>>> 9f823e2 (feat(solo): CellRanger behaviour by default on 10x geometry)
 - `--soloOutRawBarcodes Observed` writes the raw matrix with one column
   per *observed* barcode instead of one per whitelist barcode, matching
   what CellRanger's `raw_feature_bc_matrix` contains. Counts are
   unchanged; on a 200-cell run `barcodes.tsv` goes from 62 MB to 3.4 kB.
   **Not a STAR parameter**; default `Whitelist` keeps STARsolo behaviour.
->>>>>>> a8f774e (feat(solo): --soloOutRawBarcodes Observed, for a CellRanger-shaped raw matrix)
 
 - **STARsolo single-cell quantification (`--soloType`)** — the 10x
   Chromium / plate-based count-matrix pipeline, ported from STAR and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Sections commonly used: Features, Bug fixes, Other changes.
 ### Features
 
 <<<<<<< HEAD
+<<<<<<< HEAD
 - **CLI and output parity: SAM/SJ/read-input knobs and the STAR limit
   surface** — 30 further STAR 2.7.11b parameters. (`--outSAMorder` came from #145.)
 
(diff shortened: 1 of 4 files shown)


BenjaminDEMAILLE:

|---|---|---|---|

### What survives

### What does not survive

### How it happened

BenjaminDEMAILLE:

|---|---|---|---|

BenjaminDEMAILLE:
## Human data: the +0.03% is a yeast artefact, and the real gap is introns

The 20 000-read yeast fixture is the only dataset these PRs were measured on, and I flagged more than once that a second, differently-shaped one was needed. I ran it: **10x's own `pbmc_1k_v3`**, 20 M read pairs, `refdata-gex-GRCh38-2024-A` on both sides, `cellranger count` 10.0.0 against this stack plus #165.

### Raw matrix, against CellRanger

| build | entries | counts | vs CellRanger | entries identical |
|---|---|---|---|---|
| **CellRanger 10.0.0** | 3 547 245 | 9 187 379 | — | — |
| rustar, `--soloFeatures Gene` (our default) | 2 279 624 | 6 382 961 | **−30.5%** | 52.9% |
| rustar, `--soloFeatures GeneFull` | 3 544 768 | 9 039 161 | **−1.6%** | **91.8%** |

**Our default is 30.5% below CellRanger on real human data.** The +0.03% I measured on yeast is correct for that fixture and does not generalise at all — yeast has almost no introns, so exonic-only counting happened to agree. On human, CellRanger has included intronic reads by default since v7.0 and we do not.

That is one flag: `--soloFeatures GeneFull` closes it from −30.5% to −1.6%, with 91.8% of the 3.6 M entries identical cell-for-cell and gene-for-gene.

### `metrics_summary.csv` (#179), against CellRanger

The three quality metrics match **to the decimal**, which is the part of that PR I was most confident about and it holds:

| metric | CellRanger | rustar |
|---|---|---|
| Q30 Bases in Barcode | 94.2% | **94.2%** |
| Q30 Bases in RNA Read | 90.0% | **90.0%** |
| Q30 Bases in UMI | 92.8% | **92.8%** |
| Reads Mapped Confidently to Intergenic Regions | 3.7% | **3.7%** |
| Sequencing Saturation | 42.0% | 42.1% |
| Valid Barcodes | 97.3% | 97.5% |
| Reads Mapped Confidently to Intronic Regions | 31.0% | 30.6% |
| Reads Mapped Antisense to Gene | 7.9% | 7.7% |

**`Sequencing Saturation` was the metric I said I was least sure of.** On yeast it read 12.7% against CellRanger's 7.4% and I could not explain it. On human it is 42.1% against 42.0%. The yeast disagreement was a property of a fixture where no two reads share a UMI, not a defect in the formula. `DIVERGENCE.md` §3.4 needs updating to say that, and I will.

The metrics that remain far apart — `Reads Mapped Confidently to Transcriptome` 81.3% vs 55.2%, `Total Genes Detected` 23 893 vs 19 531, `Median UMI Counts per Cell` 6 353 vs 4 555 — are all downstream of the same intron difference, not separate problems.

### What I think this means for the stack

Nothing here invalidates #173–#179: the rules they implement are still the rules, and the yeast numbers in their descriptions are still what that fixture produces. What changes is the headline. **"Within 0.03% of CellRanger" is true of one small yeast fixture and false of real human data**, and I would rather say so here than have a reviewer find it.

The concrete consequence is that **`--soloFeatures GeneFull` belongs in `CELLRANGER_DEFAULTS` in #176**, alongside the five flags already there. It is a one-line change and I have not made it, because it changes default output again and #176 is already asking for sign-off on that. Your call; the measurement is above.

PR #184, opened 2026-07-31
solo: count intronic reads by default on 10x (closes 30 of the 32 points vs CellRanger, stacked on #182)

BenjaminDEMAILLE:
**Stacked on #182 → #179 → #178 → #176 → #175 → #174 → #173.** Merge in that order.

**This changes default counts on 10x runs.** First line, not last, like #176 — and it is the single largest term in the CellRanger gap.

Item 1 of the revised list in #181, and the one thing measurement showed I had backwards.

## The finding

Every number in this stack was measured on a 20 000-read yeast fixture. I flagged repeatedly that a second, differently-shaped dataset was needed. I ran it: **10x's own `pbmc_1k_v3`**, 20 M read pairs, `refdata-gex-GRCh38-2024-A` on both sides, against `cellranger count` 10.0.0.

| `--soloFeatures` | counts | vs CellRanger | entries identical |
|---|---|---|---|
| CellRanger 10.0.0 | 9 187 379 | — | — |
| `Gene` (STARsolo default, ours until now) | 6 382 961 | **−30.5%** | 52.9% |
| **`GeneFull`** | 9 039 161 | **−1.6%** | **91.8%** of 3.6 M |

CellRanger has counted intronic reads toward the gene since **v7.0**. STARsolo's `Gene` is exonic-only. On human data intronic reads are 31% of the library, so the two are 30% apart before any other flag is considered.

## Why the yeast fixture said the opposite

On yeast, `GeneFull` produced *fewer* counts than `Gene` — 14 887 against 15 439 — and I wrote in #181 that the intron decision therefore needed a dataset with real intronic signal before it was worth taking. That was the right instinct and the wrong conclusion to leave standing: yeast has almost no introns, and its genes are dense enough to overlap, so gene-body counting turns unambiguous exonic reads into ambiguous ones and drops them. The fixture was measuring its own genome, not the rule.

This is also why the stack's headline number needs reading carefully. **"Within 0.03% of CellRanger" is true of that yeast fixture and false of human data**, where the same build is 30.5% low. I have said so on #176, #178 and #179 as well.

## What this does

One entry in `CELLRANGER_DEFAULTS`, applied by the same 10x detection as #176: `--soloFeatures` defaults to `GeneFull` instead of `Gene`. Naming the flag explicitly still wins, so `--soloFeatures Gene` gets STARsolo's exonic-only counting on 10x data, and non-10x geometry is untouched.

## What it does not fix

`GeneFull` leaves us **1.6%** below CellRanger, not at parity. The next term is the MAPQ 255 rule (stage 3 of #181): CellRanger reports 93.5% of reads confidently mapped against our 88.3%, and that is unimplemented. 8.2% of matrix entries still differ. I am not claiming this closes the gap, only that it closes the largest part of it.

## Verification

- `ten_x_geometry_defaults_to_cellranger_behaviour` — extended to assert `GeneFull`
- `an_explicit_solo_features_beats_the_10x_intron_default` — `--soloFeatures Gene` survives, and the other CellRanger defaults still apply
- `non_10x_geometry_keeps_starsolo_defaults` — unchanged, 12-base barcode changes nothing

Gate: 577 lib + 27 integration tests, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, all green.

Recorded in `DIVERGENCE.md` §1.3 with the measurement. Like #176 it is a change of default output and needs sign-off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,12 @@ Sections commonly used: Features, Bug fixes, Other changes.
 
 - Read names are cut at `--readNameSeparator` (default `/`), as STAR does. A
   read named `foo/1` was previously emitted as `foo/1` where STAR emits `foo`.
+- On 10x geometry (`CB_UMI_Simple`, a whitelist, 16 bp CB, 10 or 12 bp
+  UMI), the five CellRanger-matching flags now **default** to their
+  CellRanger values. Any flag named on the command line wins, and the
+  substitution is logged. **This changes default output on 10x runs** and
+  diverges from STARsolo; see `DIVERGENCE.md` §1.3.
+
 - `--soloOutRawBarcodes Observed` writes the raw matrix with one column
   per *observed* barcode instead of one per whitelist barcode, matching
   what CellRanger's `raw_feature_bc_matrix` contains. Counts are
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,11 @@ Sections commonly used: Features, Bug fixes, Other changes.
 
 - Read names are cut at `--readNameSeparator` (default `/`), as STAR does. A
   read named `foo/1` was previously emitted as `foo/1` where STAR emits `foo`.
+- `--soloOutRawBarcodes Observed` writes the raw matrix with one column
+  per *observed* barcode instead of one per whitelist barcode, matching
+  what CellRanger's `raw_feature_bc_matrix` contains. Counts are
+  unchanged; on a 200-cell run `barcodes.tsv` goes from 62 MB to 3.4 kB.
+  **Not a STAR parameter**; default `Whitelist` keeps STARsolo behaviour.
 
 - **STARsolo single-cell quantification (`--soloType`)** — the 10x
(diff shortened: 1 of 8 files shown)


BenjaminDEMAILLE:
End-to-end confirmation on the human data, from a bare 10x command line with no flags named:

```
10x geometry detected (CB 16 + UMI 12 with a whitelist): defaulting to CellRanger
behaviour [CellRanger4, 30, 1MM_multi_Nbase_pseudocounts, MultiGeneUMI_CR, 1MM_CR,
CellRanger, GeneFull]
```

| | entries | counts | vs CellRanger |
|---|---|---|---|
| CellRanger 10.0.0 | 3 547 245 | 9 187 379 | — |
| this branch, no flags | 3 558 702 | 9 053 629 | −1.46% |

One thing to read carefully: this branch does **not** carry #165, so it sits at −1.46% rather than the −1.6% in the description, which was measured with #165 applied. #165 tightens barcode correction and therefore removes counts. The two figures are consistent, and #165 remains the merge-order dependency for the whole stack.

PR #214, opened 2026-08-06
feat(io): bound coordinate-sort memory with an external merge

alejandrogzi:
Makes `--outSAMtype BAM SortedByCoordinate` spill to disk instead of buffering the whole output in RAM.

- Coordinate sort now fills a `--limitBAMsortRAM` buffer, spills sorted runs beside the output, and k-way merges them on finish; peak memory is flat in output size.
- `--limitBAMsortRAM 0` now means 512 MiB instead of unlimited; values above 0 spill above the limit rather than aborting.
- Runs beyond 64 are merged in balanced passes so a small budget cannot exhaust file descriptors (`EMFILE`).
- Output is byte-identical to the previous in-memory sort; coordinate ties keep input order (stable merge on run index), verified by a spill-vs-unbounded test.

Behavior change vs. before: the default bounded budget replaces the previous unlimited buffering.

Tests: `cargo test --release` green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -166,6 +166,20 @@ Sections commonly used: Features, Bug fixes, Other changes.
 - **STARsolo `features.tsv` column 2 now emits the GTF `gene_name`**
   (symbol), with the STAR gene_id fallback, instead of duplicating the
   gene_id.
+- **`--outSAMtype BAM SortedByCoordinate` no longer buffers the whole
+  output in RAM.** The coordinate sort is now external: it fills a
+  `--limitBAMsortRAM` buffer, spills sorted runs beside the output, and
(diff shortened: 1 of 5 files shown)


BenjaminDEMAILLE:
Note from a pass over the PR queue: this branch is behind `main` and its green checks predate a toolchain bump, so they are stale rather than current. Clippy 1.98 added two pedantic lints that fire on code this branch still carries (fixed on `main` in #236), and CI runs `cargo clippy --all-targets -- -D warnings`, so a fresh run would fail on files this PR does not touch.

A rebase on `main` clears it. The branch lives on your fork, so it needs to come from you; every other PR in the queue that I could reach is now rebased and genuinely green.

PR #235, opened 2026-08-23
chore(ci): extend dependabot to docs/ npm and Docker, group minor/patch bumps

BenjaminDEMAILLE:
## Context

`.github/dependabot.yml` already existed and works (see the open `chore(deps)` PRs), but it watched only two ecosystems.

## Gaps closed

- **`docs/` (npm/pnpm)** — the Astro Starlight site (`docs/package.json`, `docs/pnpm-lock.yaml`: `astro`, `@astrojs/starlight`, `sharp`) was never updated.
- **Docker** — neither Dockerfile stage (`rust:1-bookworm`, `debian:bookworm-slim`) was tracked.

## Noise reduction

Minor and patch updates are grouped into one PR per ecosystem per week. Major bumps deliberately stay ungrouped: they need a real review, and a grouped major is hard to bisect. `open-pull-requests-limit` is 5 per ecosystem (3 for Docker).

Per-ecosystem commit prefixes so the history stays readable: `chore(deps)` for cargo, `chore(ci)` for actions, `chore(docs)` for the site, `chore(docker)` for base images.

## Verification

YAML parses and every group declares `patterns: ["*"]` plus explicit `update-types`. Config only: no source, workflow, or build change. GitHub validates the file on merge and reports problems in Insights > Dependency graph > Dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,12 +1,13 @@
 version: 2
 updates:
-  # Rust crates. Minor and patch bumps arrive as one grouped PR per week;
-  # major bumps stay separate because they need a real review.
+  # Rust crates. Minor and patch bumps arrive as one grouped PR per month;
+  # major bumps stay separate because they need a real review, and queue
+  # behind it (one open version-update PR per ecosystem at a time).
   - package-ecosystem: "cargo"
     directory: "/"
     schedule:
-      interval: "weekly"
-    open-pull-requests-limit: 5
+      interval: "monthly"
+    open-pull-requests-limit: 1
     commit-message:
       prefix: "chore(deps)"
     groups:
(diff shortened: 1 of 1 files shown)


Zethson (commented):
I'd even be down for monthly & an open PR limit of like 1 to drastically reduce noise.

BenjaminDEMAILLE:
Done in 378a74f: `monthly` and `open-pull-requests-limit: 1` on all four ecosystems.

Two consequences worth having on record:

- Security updates are unaffected. Per the [options reference](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference), security update PRs are exempt from `open-pull-requests-limit` and do not count toward it, so a vulnerable dependency still gets a PR immediately rather than waiting for the monthly window.
- Major bumps now queue. The limit counts all open version-update PRs for an ecosystem, so with the grouped minor/patch PR open, a major sits until that one is merged or closed. That is the intended trade for the quiet, just not invisible.

Rebased on `main` so the clippy 1.98 fix (#236) is included; CI should be green now.

PR #237, opened 2026-08-25
feat: output `anndata`

### LINKED ISSUE #1: Output `anndata` files
I'd like to tackle this once I'm back from vacation.

I need to read [the STARSolo manual](https://github.com/6c512b3b-fa9d-4cd7-b46f-d4b9b4b11aa6) closely to understand what can be handled cleanly and what can't but I think this is a good medium-term goal to output all cell/feature alignable innformation in `anndata`.  Long-term, having some sort of "aligner schema" at a high level where we output information for identifying how a count matrix was generated would be great but is a stretch.

For `anndata`, there are now readers in most bioinformatics-relevant languages I can think of and [the format is well-documented](https://anndata.readthedocs.io/en/latest/fileformat-prose.html): https://github.com/kaizhang/anndata-rs, https://github.com/ilan-gold/anndata.js, https://github.com/scverse/anndataR, and maybe others I'm missing (Julia?).

Ideally we would output zarr because it is both [the fastest (ideally with sharding + v3)](https://github.com/zarrs/zarr_benchmarks) and best cross-language (with the main outlier being R, where there is ongoing work: https://github.com/scverse/anndataR/pull/190).  Relatedly, hdf5 is not cloud friendly (i.e., browser or local + no prospect for support in JavaScript unlike in R for zarr) and is generally slower from what I have observed (in rust + python at least, multithreading in zarr is the default and works well).  However, providing options for both is probably realistic.

The other thing about zarr is that [`SpatialData`](https://github.com/scverse/spatialdata) only supports zarr.

ilan-gold:
Fixes #1 

See https://github.com/scverse/rustar-aligner/issues/1#issuecomment-5232297955 for the plan / approach.

Currently dependent on the bug highlighted in https://github.com/scverse/anndata-rs/pull/44 which may not be a bug but instead something that should be changed in python.

Todos:

- [ ] Resolve the `obsm` `dataframe` mess
- [ ] Clean up some of the AI output
- [ ] Relatedly, decide on how extras should be structured
- [ ] Check speed


BenjaminDEMAILLE:
Heads-up on the conflict state, from a pass over the whole PR queue.

Two of the three conflicts against current `main` are mechanical and I verified a resolution locally:

- `src/solo/mod.rs`: `pub mod adata;` and `pub mod cell_reads;` are both additions, so both stay.
- `src/junction/sj_output.rs`: `main` still inlines the surviving-junction filter and sort inside `write_sj_lines`, while this branch factored the same logic into `sorted_junctions(params)`. The helper is equivalent (same filter, same sort key), so this branch's version wins and `main`'s inline block goes.

The third is not mechanical and is yours to decide: `src/solo/count.rs` has two ~110-line hunks where this branch rewrote the matrix writer to take a `cells` slice, and `main` reworked the same function's record folding (1MM_multi resolution off `exact_count_snapshot`, `std::mem::take` of the recorder). Picking one shape over the other changes what the anndata writer receives, so I left it alone rather than guessing.

No commits were pushed to this branch.

Issue #48, opened 2026-05-12
`Log.final.out` folds all unmapped reads into `too short`; `other` bucket always 0


pinin4fjords:
## Summary

`Log.final.out` reports two categories of unmapped reads — `Number of reads unmapped: too short` and `Number of reads unmapped: other` — using STAR's exact field labels. rustar folds both buckets into `too short` and always reports `other = 0`. Total unmapped count is preserved, but a downstream tool that parses the two fields separately (MultiQC's STAR module, custom QC dashboards) gets an inflated `too short` percentage and a zero `other` percentage.

On `WT_REP2` (nf-core/rnaseq test profile):

| Field | STAR 2.7.11b | rustar 0.1.0 |
|---|---:|---:|
| `Number of reads unmapped: too many mismatches` | 0       | 0 |
| `Number of reads unmapped: too short`           | 1 540 (3.11 %) | 4 193 (8.46 %) |
| `Number of reads unmapped: other`               | 2 656 (5.36 %) | 0 (0.00 %) |
| **Total unmapped**                              | **4 196**      | **4 193**     |

The totals match within RNG / tie-breaking tolerance — rustar isn't dropping reads. It's the bucketing that's wrong: every read STAR would label `other` (no seeds / no clusters / failed-stitch reason) is being mislabelled as `too short` (post-stitch length / score / match filter).

## STAR reference behaviour

STAR's bucketing (per [`source/ReadAlign_assignAlignToWindow.cpp`](https://github.com/alexdobin/STAR/blob/master/source/ReadAlign_assignAlignToWindow.cpp) and the `nUnmapped` arrays in `source/Stats.cpp`):

- `too short`: a transcript was generated but filtered out by score / match-count / read-length-coverage filters (`outFilterScoreMin*`, `outFilterMatchNmin*`).
- `other`: no seeds, no clusters, no transcript — read couldn't be aligned at all.
- `too many mismatches`: filtered by `outFilterMismatchN*`.

rustar's `src/stats.rs::UnmappedReason` enum has all four variants:

```rust
pub enum UnmappedReason {
    Other,
    TooShort,
    TooManyMismatches,
    TooManyLoci,
}
```

So the categorization exists in the code. The bug is at the **classification call sites** — wherever an unmapped read is recorded, the code is calling `record_unmapped_reason(UnmappedReason::TooShort)` instead of `UnmappedReason::Other` for reads that have no alignments at all.

## Reproducer

```bash
#!/usr/bin/env bash
[...]

Issue #31, opened 2026-05-12
Multi-mapper NH cap differs from STAR (NH up to 20 vs STAR's 7); possibly missing `--outFilterMultimapScoreRange`


pinin4fjords:
## Summary

With identical `--outFilterMultimapNmax 20` on the same paired-end inputs, rustar emits multi-mappers with NH values up to 20, while STAR caps at NH=6 on the same data. The differing tail isn't fully accounted for by tie-breaking — rustar is keeping additional, lower-scoring secondary alignments that STAR filters out.

Could be intentional, but the visible pattern (a long tail of NH=7..20 records that STAR doesn't produce) looks like a missing or differently-configured score-range filter.

## STAR reference behaviour

STAR applies **two** caps on multi-mapper count, not one. From [`source/Parameters.cpp:141-142`](https://github.com/alexdobin/STAR/blob/master/source/Parameters.cpp#L141-L142):

```cpp
parArray.push_back(new ParameterInfoScalar <uint>     (-1, -1, "outFilterMultimapNmax",      &outFilterMultimapNmax));
parArray.push_back(new ParameterInfoScalar <intScore> (-1, -1, "outFilterMultimapScoreRange", &outFilterMultimapScoreRange));
```

- `--outFilterMultimapNmax` (default 10): hard ceiling on number of loci a read can map to.
- `--outFilterMultimapScoreRange` (default **1**): max score gap between best and a kept multi-mapper. Alignments with score < best_score - scoreRange are dropped before the Nmax cap is applied.

So with `--outFilterMultimapNmax 20` and the default `--outFilterMultimapScoreRange 1`, STAR keeps only loci within 1 score unit of the best, then caps at 20. The empirical NH<=6 tail is a consequence of the score-range filter trimming much earlier than the count filter.

rustar's `--help` output doesn't list a `--outFilterMultimapScoreRange` option, consistent with the filter not being implemented yet.

## Reproducer

```bash
#!/usr/bin/env bash
set -euo pipefail
mkdir -p /tmp/rustar-mre-31 && cd /tmp/rustar-mre-31

BASE=https://raw.githubusercontent.com/nf-core/test-datasets/626c8fab639062eade4b10747e919341cbf9b41a
curl -fsLO $BASE/reference/genome.fasta
curl -fsL  $BASE/reference/genes_with_empty_tid.gtf.gz | gunzip -c > genes.gtf
[...]

PR #249, opened 2026-08-27
ci: opt-in large-dataset differential against STAR

### LINKED ISSUE #244: Enable middle/larger scale dataset tests
Let's have an Actions workflow where we provision custom runners via scverse AWS EC2 to trigger bigger tests. Maybe dispatching first and not on every PR or via a label and then we go from there

BenjaminDEMAILLE:
Closes #244.

## Why

The suites in `tests/` use a synthetic micro-genome and finish in seconds. That is why the two bugs reported from nf-core/rnaseq data (#31, #48) were invisible to them: both are about aggregate behaviour over tens of thousands of reads, not about one alignment.

## What runs

`test/nfcore_diff.py` fetches the nf-core/rnaseq test dataset from public URLs (50 000 paired reads, *S. cerevisiae* chrI plus the GFP transgene — the same fixture the reporter used), builds both indexes, runs STAR and rustar-aligner, and compares:

- uniquely mapped, multi-mapped, `unmapped: too short`, `unmapped: other`, each as a fraction of input reads;
- the deepest multimapper (max NH), as a ratio.

Thresholds are the measured gaps plus headroom, so today's state passes and a regression trips them. `--report-only` prints the comparison and exits 0, which is what to use while a difference is being investigated.

## Checked against a known regression

Run on `main` right now, the script fails on exactly the two buckets #247 fixes, and reports #31's depth gap:

```
metric                                         STAR     rustar      delta
Uniquely mapped reads number                  41691      41665    0.052%
Number of reads mapped to multiple loci         934        961    0.054%
Number of reads unmapped: too short            3766       7374    7.216%
Number of reads unmapped: other                3609          0    7.218%
deepest multimapper (NH)                         14         20      1.43x

FAILED:
  Number of reads unmapped: too short: 7.216% > 1.000%
  Number of reads unmapped: other: 7.218% > 1.000%
```

With #247 merged those two lines read 3778 and 3596, inside the thresholds.

## How it is triggered

Opt-in, never on every push: manual dispatch, or the `large-tests` label on a pull request. The dispatch form takes a runner label, so it can be aimed at a scverse AWS runner without editing the workflow — that is the part of the issue I cannot do from here, since it needs the runner group to exist.

STAR comes from Ubuntu's `rna-star` package; building it from source would triple the job runtime without changing what is measured.

One note carried into the script: `--readFilesCommand zcat` yields zero reads on macOS for **both** aligners, silently, so the fixture is decompressed up front rather than piped.

### DIFF
--- a/.github/workflows/large-dataset.yml
+++ b/.github/workflows/large-dataset.yml
@@ -1,0 +1,71 @@
+name: Large-dataset differential
+
+# Deliberately not on every push: this job fetches a dataset, builds two
+# indexes and runs two aligners. It is opt-in, either by hand or by putting the
+# `large-tests` label on a pull request.
+on:
+  workflow_dispatch:
+    inputs:
+      runner:
+        description: "Runner label (e.g. ubuntu-latest, or a scverse AWS runner group)"
+        required: false
+        default: "ubuntu-latest"
+      report_only:
+        description: "Report the comparison without failing the job"
+        type: boolean
+        required: false
+        default: false
+  pull_request:
+    types: [labeled, synchronize]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  nfcore-diff:
+    name: nf-core/rnaseq differential against STAR
+    # On a pull request this runs only when the `large-tests` label is present,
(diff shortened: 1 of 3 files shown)


Zethson:
We have a specific vision for this and I had assigned this to myself. If you work on an issue that isn't assigned to you, please talk to us first. We don't want a one-shot LLM PR here, please.

PR #250, opened 2026-08-27
ci: opt-in benchmark workflow, micro and end-to-end

### LINKED ISSUE #245: Enable performance (here speed) benchmarking
Let's try to reuse our scverse benchmarking setup but ideally with our scverse AWS resources

BenjaminDEMAILLE:
Closes #245. Stacked on #248 (the divan harness), which it uses.

Two jobs, both opt-in the way #249's large-dataset job is (manual dispatch, or a `benchmark` label on a pull request), because a wall time from a busy shared runner is worse than no wall time.

## `micro`

Runs `cargo bench --bench hot_paths`, optionally against a baseline ref, and writes a table into the job summary:

| benchmark | this ref | baseline | change |
|---|---|---|---|
| gene_annotation_build/20000 | 12.42 ms | 12.42 ms | +0.0% |
| gene_overlap_query/100 | 98.70 ns | 70.00 ns | +41.0% ⚠️ |

A baseline older than the harness has no benches to run; the job says so rather than failing.

## `end-to-end`

Times indexing and alignment **separately** for both aligners on the nf-core fixture, with peak RSS, because they are different questions: indexing is dominated by suffix-array construction, alignment by the seed and stitch loops. Deliberately without thresholds — it is a data point, and only comparable within one machine.

## The parser was checked both ways

`test/bench_report.py` turns divan's box-drawing tree into the table. My first version silently lost the group names and reported rows called `100` and `2000`; the fix keys them as `gene_overlap_query/100`. It is also checked against a perturbed baseline: with one row moved from 98.7 ns to 70 ns in the baseline file, that row is flagged at +41.0% and every other row reads +0.0%.

## What I could not do from here

The issue asks to reuse the scverse benchmarking setup on scverse AWS resources. The runner label is an input with `ubuntu-latest` as the default, so pointing this at a runner group is a one-word change once that group exists — but I cannot create it, and I did not guess at its name.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,111 @@
+name: Benchmarks
+
+# Benchmarks are opt-in for the same reason the large-dataset job is: they are
+# long, and a number measured on a noisy shared runner is worse than no number.
+on:
+  workflow_dispatch:
+    inputs:
+      runner:
+        description: "Runner label (e.g. ubuntu-latest, or a scverse AWS runner group)"
+        required: false
+        default: "ubuntu-latest"
+      baseline:
+        description: "Git ref to compare against (empty = no comparison)"
+        required: false
+        default: "main"
+  pull_request:
+    types: [labeled, synchronize]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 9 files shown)


Zethson:
We have a specific vision for this and I had assigned this to @flying-sheep. If you work on an issue that isn't assigned to you, please talk to us first. We don't want a one-shot LLM PR here, please.

